PR #8, opened 2015-04-03
Ignore .gem files in gitignore

ihmccreery:
@plribeiro3000 Sound good?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.gem
 /.bundle/
 /.yardoc
 /Gemfile.lock


plribeiro3000:
Hmmm. I guess you don't need that since [this line](https://github.com/fog/fog-google/pull/8/files#diff-a084b794bc0759e7a6b77810e01874f2R8) should take care of it. Why do you have a `.gem` file in the root folder? It should be under `pkg`.

ihmccreery:
Bah, no, I used `gem build` and `gem push`.  I'll use `rake release` in the future.  I checked and saw that this line is in `fog/.gitignore`, so I thought it was good.

PR #14, opened 2015-04-06
Remove `turn` gem from testing due to deprecation

ihmccreery:


ihmccreery:
@plribeiro3000 Can you please confirm this is an appropriate change?  Per [Turn no longer being maintained](https://github.com/turn-project/turn#turn-is-no-longer-being-maintained).

plribeiro3000:
@ihmccreery Interesting. Didn't know about it.
I guess its because the usage is so simple that i did not hit any bug. I think we can remove it.

I will search for another solution when i find a time. Blank and white terminals are a nightmare for me. ehhehe

PR #21, opened 2015-04-21
Repair rescue logic for get images

### LINKED ISSUE #20: Error in Ruby's rescue clause
In the following file/commit/line:

https://github.com/fog/fog-google/commit/1976f5c62027042311809e5d17de36bfec833dcb#commitcomment-10822546

Causes (at least) the following error:

    /.../ruby/2.2.0/gems/fog-google-0.0.3/lib/fog/google/models/compute/images.rb:67:in `rescue in get': class or module required for rescue clause (TypeError)
        from /.../ruby/2.2.0/gems/fog-google-0.0.3/lib/fog/google/models/compute/images.rb:47:in `get'
        from /.../ruby/2.2.0/gems/fog-google-0.0.3/lib/fog/google/requests/compute/insert_disk.rb:77:in `insert_disk'
        from /.../ruby/2.2.0/gems/fog-google-0.0.3/lib/fog/google/models/compute/disk.rb:40:in `save'
        from /.../ruby/2.2.0/gems/fog-core-1.30.0/lib/fog/core/collection.rb:51:in `create'
        from ...

ihmccreery:
Revert logic from fog/fog@39188c331cbe5e51169f5dcd12694a998f5502ba, make it consistent with desired functionality; closes #20.

@plribeiro3000 Can you please review?

### DIFF
--- a/lib/fog/google/models/compute/images.rb
+++ b/lib/fog/google/models/compute/images.rb
@@ -57,15 +57,8 @@ module Fog
               break
             end
           end
-
-          # If it wasn't found in any project, raise
           return nil if data.nil?
-
           new(data)
-        
-        rescue Fog::Errors::NotFound.new(
-          "Unable to find the image #{identity} in the following projects: #{all_projects.join(', ')}")
-
         end
       end
     end


plribeiro3000:
Hmmm. I think that perhaps this rescues does not make sense at all.
Shouldn´t [this](https://github.com/fog/fog-google/pull/21/files#diff-54f066af2ced1b900d56882c80ec2d3bR54) rescue catch all the exceptions?

ihmccreery:
Yes, it should.  I was thinking that it'd be better to be consistent with other methods, but upon further consideration, I've taken it out.

plribeiro3000:
Somehow, it seems the original intentio was to use `raise`, not `rescue`.

```ruby
raise Fog::Errors::NotFound.new(
  "Unable to find the image #{identity} in the following projects: #{all_projects.join(', ')}")
```

Does it make sense related to the rest of the project?

ihmccreery:
Yeah, from what I've seen, I think this was a copy-paste error combined with a functionality change that didn't get covered in tests.

Per the standard `collection_tests` [here](https://github.com/fog/fog-google/blob/master/tests/helpers/collection_helper.rb#L91), if we can't find a image, we should return `nil`, not raise an error.  The original commit that caused this problem was putting images under this test.

plribeiro3000:
Awesome. Thank you for the clarification! This is good to go!

PR #24, opened 2015-04-22
Fix BackendServices functionality to reflect standard functionality

ihmccreery:
- `get` should return nil when finding nothing
- fix `create` and `destroy`
- fix tests

### DIFF
--- a/lib/fog/google/models/compute/backend_services.rb
+++ b/lib/fog/google/models/compute/backend_services.rb
@@ -13,8 +13,12 @@ module Fog
         end
 
         def get(identity)
+          response = nil
           response = service.get_backend_service(identity)
-          new(response.body) unless response.nil?
+          return nil if response.nil?
+          new(response.body)
+        rescue Fog::Errors::NotFound
+          nil
         end
       end
     end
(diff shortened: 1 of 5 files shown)


ihmccreery:
@plribeiro3000 can I get a :+1:?

lib/fog/google/models/compute/backend_services.rb
```diff
@@ -13,8 +13,12 @@ def all(filters={})
         end
 
         def get(identity)
+          resonse = nil
```
plribeiro3000: Why this?

plribeiro3000:
@ihmccreery sure. made a small comment. It seems to be a typo but i'm not sure the real intent behind it if it is corrected since the next line overwrite it anyway.

ihmccreery:
Good catch.  That's actually a typo [here](https://github.com/fog/fog-google/commit/1976f5c62027042311809e5d17de36bfec833dcb#L16) as well, which I'll fix.

ihmccreery:
*This was merged.*  Still getting used to the rebase, sorry.

plribeiro3000:
Don't worry. it happens to us all. =)

PR #26, opened 2015-04-25
Get IP address by name

Temikus:
Adding a get_by_name method to allow fetching of an IP address by name.

P.S. I needed to add a line to contributors, since this change is a part of my work at Google.

CC/ @erjohnso 
This change is made to later implement this feature in vagrant-google.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -39,3 +39,4 @@
 * Timur Alperovich <timur.alperovich@gmail.com>
 * unknown <bturner_2@pibuk-lp71.pibenchmark.com>
 * Wesley Beary <geemus@gmail.com>
+* Google Inc.
--- a/lib/fog/google/models/compute/addresses.rb
+++ b/lib/fog/google/models/compute/addresses.rb
@@ -34,6 +34,19 @@ module Fog
           return nil if address.empty?
           new(address.first['addresses'].first)
         end
+
+        def get_by_name(ip_name)
+          names = service.list_aggregated_addresses(:filter => "name eq .*#{ip_name}").body['items']
+          name = names.each_value.select { |region| region.key?('addresses') }
+
+          return nil if name.empty?
+          new(name.first['addresses'].first)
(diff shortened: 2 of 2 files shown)


ihmccreery:
Thanks @Temikus.

I'm concerned about this line:

```ruby
return nil if address.empty?
```

Should it be:

```ruby
return nil if name.empty?
```

How much have you tested it?  I would like to ask you to write tests, but I realize that the state of testing in this repo is such that that's not really reasonable.  Otherwise, the change looks reasonable.

Temikus:
Yep, you're right. Now I see I need to test this better.
I'll update as soon as I make sure it's working.

Temikus:
Fixed and added a new method (get_by_ip_address_or_name) for easy Addressretrieval.
Tested with latest git snapshots of fog-core and fog:
```
  gem 'fog-core', git: "https://github.com/fog/fog-core"
  gem 'fog-google', path: "."
  gem 'fog', git: "https://github.com/fog/fog"
```

Let me know if I need to change anything.

ihmccreery:
@Temikus Thanks so much.

Can't `get_by_ip_address_or_name` just call out to the other two methods, as opposed to duplicating all of the business logic?  I feel like it should be a one-liner, something like:

```ruby
def get_by_ip_address_or_name(ip_address_or_name):
    get_by_ip_address(ip_address_or_name) or get_by_name(ip_address_or_name)
end
```

And, please squash your work into a single commit, and rebase to master if necessary.

Temikus:
Squashed and re-written to reuse business logic.

ihmccreery:
This [has been merged](https://github.com/fog/fog-google/commit/f46634c0de8bac65ec5fc36a9b9ae99e3038d4ed); I don't know why GitHub isn't marking it as such.

ihmccreery:
Oh, it's because it wasn't rebased to master, so I rebased it and thus changed the commit hash.  Sorry 'bout that.  Still learning.

PR #28, opened 2015-04-28
Adding bitnami-launchpad global project

migmartri:
We offer an up to date public catalog of open source application images on GCE.

The community might be interested in accessing to those images via fog.

Thank you for your work on this library! :)

```
gcloud compute images list --project bitnami-launchpad
NAME                                                    PROJECT           ALIAS              DEPRECATED STATUS
bitnami-abantecart-1-2-1-0-linux-debian-7-x86-64        bitnami-launchpad                               READY
bitnami-activemq-5-11-1-0-linux-debian-7-x86-64         bitnami-launchpad                               READY
bitnami-akeneo-1-3-9-0-linux-debian-7-x86-64            bitnami-launchpad                               READY
....
```

### DIFF
--- a/lib/fog/google/models/compute/images.rb
+++ b/lib/fog/google/models/compute/images.rb
@@ -18,7 +18,8 @@ module Fog
           'opensuse-cloud',
           'rhel-cloud',
           'suse-cloud',
-          'ubuntu-os-cloud'
+          'ubuntu-os-cloud',
+          'bitnami-launchpad'
         ]
 
         def all


ihmccreery:
@migmartri Thanks for this, but I don't think this is viable.  There are many projects, both "official" and "unofficial," that provide public catalogs of images, and to include any beyond the standard set is probably not maintainable.

May I suggest an alternative approach?  It would be great if we could allow users to specify other projects that they would like to include in their images search.  That way, if someone wanted to use `bitnami-launchpad` images, they could do so.

migmartri:
I understand, I'll come back with an alternative implementation.

Thanks,

ihmccreery:
Awesome, thanks so much!  Let me know if you have any questions, etc.

I'm going to go ahead and close this PR; feel free to open a new one with the alternative implementation.

PR #29, opened 2015-04-29
Override current project ID on images list/get operations

migmartri:
Following the same behaviour than the official gcloud CLI (attached image), now it is possible to pass a project option to override the target project ID.

* If no project parameter is passed, the project related to the service is used.
* The current or specific project images takes preference on the returned list. 
* This override affects to both list and describe API calls.

selection_378


migmartri:
Just noticed that this solution does not propagate correctly to actions like insert_disk in which image = images.get(image_name) get's called.

I will be back :)

PR #30, opened 2015-04-29
google_extra_global_projects on initialization for compute

migmartri:
Added the option for setting extra global projects during initialization.

My previous approach did not fully work if you want to rely on these extra global projects for disk creation for example. https://github.com/fog/fog-google/pull/29

### DIFF
--- a/lib/fog/google/compute.rb
+++ b/lib/fog/google/compute.rb
@@ -220,9 +220,11 @@ module Fog
 
       class Mock
         include Fog::Google::Shared
+        attr_reader :extra_global_projects
 
         def initialize(options)
           shared_initialize(options[:google_project], GOOGLE_COMPUTE_API_VERSION, GOOGLE_COMPUTE_BASE_URL)
+          @extra_global_projects = options[:google_extra_global_projects] || []
         end
 
         def self.data(api_version)
--- a/lib/fog/google/compute.rb
+++ b/lib/fog/google/compute.rb
@@ -3,7 +3,7 @@ module Fog
     class Google < Fog::Service
       requires :google_project
       recognizes :app_name, :app_version, :google_client_email, :google_key_location, :google_key_string,
-                 :google_client, :google_json_key_location, :google_json_key_string
+                 :google_client, :google_json_key_location, :google_json_key_string, :google_extra_global_projects
 
       GOOGLE_COMPUTE_API_VERSION     = 'v1'
       GOOGLE_COMPUTE_BASE_URL        = 'https://www.googleapis.com/compute/'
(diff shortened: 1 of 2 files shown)


ihmccreery:
Thanks!

Why did you choose to put `:google_extra_global_projects` in `Fog::Compute::Google` as opposed to keeping it contained within `Fog::Compute::Google::Images`?  It seems to me that variable should be in the same scope as `GLOBAL_PROJECTS`, (and the only reason you'd want to query other projects is to get at public images, right, so shouldn't it stay in the scope of `Images`?).

Otherwise, this looks good.  Once we've figured out the above issue, I'll ask you to squash the commits, and I'll test it out and merge it in.

ihmccreery:
(On the other hand, I can see the benefit of having this be something you just pass in at the beginning, when you're making the connection, and not worrying about it after.  I'm just curious if you explored the option of keeping it confined to `Images`.)

migmartri:
Hello, thanks for reviewing it.

> Why did you choose to put :google_extra_global_projects in Fog::Compute::Google as opposed to keeping it contained within Fog::Compute::Google::Images?

I wanted to follow a convention in which I only modify the fog/compute (or any other class) instance on an explicit object instantiation. This attribute will be exposed everywhere in any different operation requested (disk_insert, images.list, etc)

With my solution, the initialization looks like:

      ::Fog::Compute.new(provider: 'Google',
                         google_client: client,
                         google_project: project_id,
                         google_extra_global_projects: ['bitnami-launchpad']
                        )


Which does not require modifying/redefining the images initialize method, which (I am new contributing on fog), comes inherited from core/collection. Also, as far I can see, most of the classes follow a standard method implementation (all, get methods), I did not want to modify or add attrs_accessors there.

In the other hand, if we decide to put the modifier in the images class, I would need to (off the top of my head) do smth like:

@fog =   ::Fog::Compute.new(provider: 'Google',
                       google_client: client,
                       google_project: project_id,
               )

@fog.images.extra_global_projects = [bitnami-launchpad]

What worries me about this solution is that I am not 100% sure how the fog "metaprogramming" works on fog since it seems to rely on loading the classes when they are used. I am worried that at some point of the execution of my program, the images class gets re-instantiated and I loose the extra_global_projects instance attribute.   

In any case, you have a better understanding of the library and the patterns that must to be followed. As you can see, the explanation above is more about a preference than an strong opinion. I will be happy to discuss alternative approaches. 

Thanks,
Miguel

ihmccreery:
Yeah, I figured it was something like that.  Thanks for the thorough explanation.

```ruby
> g = Fog::Compute.new(provider: 'Google',
*                      google_extra_global_projects: ['bitnami-launchpad'])
=> ...
> g.images.map(&:name)
=> [ ... , "bitnami-abantecart-1-2-0-0-r02-linux-debian-7-x86-64", ... ]
```

Looks great!

migmartri:
Thanks! :)

PR #32, opened 2015-05-23
correct :user to :username in launch instance example

digitalronin:
As written, the instance will always be created with a username matching ENV['USER']  If you send the parameter as 'username' it works as expected.


icco:
Thanks!

Issue #36, opened 2015-05-29
Scope aliases consistency with gcloud


Temikus:
We've had a discussion with @erjohnso about service account scopes inconsistency with official `gcloud` utility in https://github.com/mitchellh/vagrant-google/pull/71.

If one wants to define a service account scope using a short name, we have the following (left is `gcloud compute`, right is fog `scope:` attribute, since `gcloud compute` aliases do not match up to API endpoints:
```
          compute-ro         compute.readonly
          compute-rw         compute
          computeaccounts-ro computeaccounts.readonly
          computeaccounts-rw computeaccounts
          logging-write      logging.write
          sql                sqlservice
          sql-admin          sqlservice.admin
          storage-full       devstorage.full_control
          storage-ro         devstorage.read_only
          storage-rw         devstorage.read_write
```
Excerpt from `gcloud compute instances create --help`:
```
          Alias              URI
          bigquery           https://www.googleapis.com/auth/bigquery
          cloud-platform     https://www.googleapis.com/auth/cloud-platform
          compute-ro         https://www.googleapis.com/auth/compute.readonly
          compute-rw         https://www.googleapis.com/auth/compute
          computeaccounts-ro https://www.googleapis.com/auth/computeaccounts.readonly
          computeaccounts-rw https://www.googleapis.com/auth/computeaccounts
          datastore          https://www.googleapis.com/auth/datastore
          logging-write      https://www.googleapis.com/auth/logging.write
          monitoring         https://www.googleapis.com/auth/monitoring
          sql                https://www.googleapis.com/auth/sqlservice
          sql-admin          https://www.googleapis.com/auth/sqlservice.admin
          storage-full       https://www.googleapis.com/auth/devstorage.full_control
          storage-ro         https://www.googleapis.com/auth/devstorage.read_only
          storage-rw         https://www.googleapis.com/auth/devstorage.read_write
          taskqueue          https://www.googleapis.com/auth/taskqueue
          userinfo-email     https://www.googleapis.com/auth/userinfo.email
```

I was thinking about implementing it by either:
1) Keeping an additional mapping of those shortcuts and if one matches - expanding them (Since it's probably a bad idea to have only `gcloud`-style mappings available in scopes, as that may break things for existing users, who expect the old-style shortcut).
or
2) Creating a separate scope_aliases: attribute which will accept only aliased parameters in the same format as gcloud, expanding them and passing them on to service_accounts: later.

Any thoughts on this?

P.S. This has already been done in libcloud:
[...]

PR #46, opened 2015-06-01
Adding support for gcloud scope aliases

Temikus:
Fixing #36. Adding gcloud compute aliases support to be consistent with official utility.

### DIFF
--- a/lib/fog/google/models/compute/server.rb
+++ b/lib/fog/google/models/compute/server.rb
@@ -31,6 +31,21 @@ module Fog
         attribute :auto_restart
         attribute :on_host_maintenance
 
+        # Security account scope aliases used by official gcloud utility
+        # List derived from 'gcloud compute instances create --help'
+        GCE_SCOPE_ALIASES = {
+              'compute-ro' => 'compute.readonly',
+              'compute-rw' => 'compute',
+              'computeaccounts-ro' => 'computeaccounts.readonly',
+              'computeaccounts-rw' => 'computeaccounts',
+              'logging-write' => 'logging.write',
+              'sql' => 'sqlservice',
+              'sql-admin' => 'sqlservice.admin',
+              'storage-full' => 'devstorage.full_control',
+              'storage-ro' => 'devstorage.read_only',
+              'storage-rw' => 'devstorage.read_write'
+            }
+
         def image_name=(args)
           Fog::Logger.deprecation("image_name= is no longer used [light_black](#{caller.first})[/]")
         end
@@ -206,6 +221,27 @@ module Fog
           return self.metadata
         end
 
+        def map_service_accounts(scope_array)
+
+          scope_array_expanded = scope_array.map do |e|
+            if GCE_SCOPE_ALIASES[e]
+              GCE_SCOPE_ALIASES[e]
+            else
+              e
+            end
+          end
+
+          scope_array_finalized = scope_array_expanded.map do |e|
+            if e.start_with?("https://")
+              e
+            else
+              "https://www.googleapis.com/auth/#{e}"
+            end
+          end
+
+          return scope_array_finalized
+        end
+
         def reload
           data = service.get_server(self.name, zone_name).body
           self.merge_attributes(data)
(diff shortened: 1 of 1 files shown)


lib/fog/google/models/compute/server.rb
```diff
@@ -206,6 +206,32 @@ def add_ssh_key username, key
...
+            'computeaccounts-rw' => 'computeaccounts',
+            'logging-write' => 'logging.write',
+            'sql' => 'sqlservice',
+            'sql-admin' => 'sqlservice.admin',
+            'storage-full' => 'devstorage.full_control',
+            'storage-ro' => 'devstorage.read_only',
+            'storage-rw' => 'devstorage.read_write'
+              }
+
+          scope_array.map! {
```
ihmccreery: This is mutating `scope_array`, (`map!` is destructive).  I'd either switch to non-destructive `map` and returning a new array, (I think this is what I'd prefer,) or entitle the method `map_service_accounts!`, and probably call it in the constructor.  Mutating class variables during `save` sounds scary to me.

Stylistically, for a multi-line block, you should use the `do end` syntax, and use brackets for inline.  See [this answer](http://stackoverflow.com/a/5587403/594530).

lib/fog/google/models/compute/server.rb
```diff
@@ -206,6 +206,32 @@ def add_ssh_key username, key
           return self.metadata
         end
 
+        def map_service_accounts(scope_array)
+          gce_scope_aliases = {
```
ihmccreery: `gce_scope_aliases` seems like it should be a constant, declared in the class scope or somewhere else.

lib/fog/google/models/compute/server.rb
```diff
@@ -206,6 +206,32 @@ def add_ssh_key username, key
...
+            'sql' => 'sqlservice',
+            'sql-admin' => 'sqlservice.admin',
+            'storage-full' => 'devstorage.full_control',
+            'storage-ro' => 'devstorage.read_only',
+            'storage-rw' => 'devstorage.read_write'
+              }
+
+          scope_array.map! {
+            |e| gce_scope_aliases[e] ? gce_scope_aliases[e] : e
+              }
```
ihmccreery: Make sure your blocks are indented and outdented properly and consistently.

ihmccreery:
Hey Artem, thanks for this work!  See inline comments.

Temikus:
Hi Ike,

Consistently outdented blocs (used previous indentation model), moved the aliases away into a constant. Squashed the commits.
Let me know if I missed something.

ihmccreery:
Hi Artem,

You're still doing a destructive operation during `save`:

> This is mutating `scope_array`, (`map!` is destructive). I'd either switch to non-destructive `map` and returning a new array, (I think this is what I'd prefer,) or entitle the `method map_service_accounts!`, and probably call it in the constructor. Mutating class variables during `save` sounds scary to me.

Unless you have a strong argument in favor of doing so, I recommend doing either a non-destructive operation, or switching to call this method during initialization.

Also,

> Stylistically, for a multi-line block, you should use the do end syntax, and use brackets for inline. See this [answer](http://stackoverflow.com/questions/5587264/do-end-vs-curly-braces-for-blocks-in-ruby/5587403#5587403).

Temikus:
Hi Ike,

Thanks for the comments!

1) About destructive `map!`
Sorry, missed that particular one, my bad. Agreed, changed to a non-destructive map.

2) About one line if's
Initially, I expanded that statement only because it was over 80 chars. Rubocop indicated that to me, however, you already had a multiline statement like this:
```
              "scopes" => service_accounts.map {
                |w| w.start_with?("https://") ? w : "https://www.googleapis.com/auth/#{w}"
              }
```
, and the unless/if block looks a little messier so wasn't sure how to proceed.
I now mapped it in a full if/else block.

Branch updated, commits squashed. Let me know if you have any comments.

Cheers,
A.

ihmccreery:
Looks good to me!  I don't really have a preference for `_ ? _ : _` versus `if`/`else` syntax, but this looks fine.

Thanks for all your work on this!

PR #55, opened 2015-06-24
added preemptible

jcdang:
https://cloud.google.com/compute/docs/instances/preemptible

### DIFF
--- a/lib/fog/google/requests/compute/insert_server.rb
+++ b/lib/fog/google/requests/compute/insert_server.rb
@@ -152,7 +152,7 @@ module Fog
           scheduling = {
             'automaticRestart' => false,
             'onHostMaintenance' => "MIGRATE",
-            'preemptible' => true
+            'preemptible' => false
           }
           if options.key? 'auto_restart'
             scheduling['automaticRestart'] = options.delete 'auto_restart'
--- a/lib/fog/google/requests/compute/insert_server.rb
+++ b/lib/fog/google/requests/compute/insert_server.rb
@@ -151,11 +151,16 @@ module Fog
 
           scheduling = {
             'automaticRestart' => false,
-            'onHostMaintenance' => "MIGRATE"
+            'onHostMaintenance' => "MIGRATE",
+            'preemptible' => true
           }
           if options.key? 'auto_restart'
             scheduling['automaticRestart'] = options.delete 'auto_restart'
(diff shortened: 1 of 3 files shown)


lib/fog/google/requests/compute/insert_server.rb
```diff
@@ -151,12 +151,17 @@ def insert_server(server_name, zone_name, options={}, *deprecated_args)
 
           scheduling = {
             'automaticRestart' => false,
-            'onHostMaintenance' => "MIGRATE"
+            'onHostMaintenance' => "MIGRATE",
+            'preemptible' => true
```
ihmccreery: It seems scary to me that we would default to preemptible VMs, if the user doesn't specify.  That's also a very large change from what the current behavior is.

ihmccreery:
@jcdang Thanks for this.  See my note above; otherwise, this looks great.  Can you confirm how you've tested it?

Temikus:
Sorry for intruding, but ditto on the scary default policy. 

What is especially concerning:
>This is a Beta release of Preemptible Instances. This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
>Additionally, Compute Engine has a finite number of available preemptible instances, so you might not be able to create them during peak usage.

This will require either overwriting the setting in dependant projects or introduce fallback behaviour against running out of available instances.

lib/fog/google/requests/compute/insert_server.rb
jcdang: Oops that's a mistake when I was testing.

jcdang:
@ihmccreery I ran the unit tests with rake and some of the integration tests with shindont +google (after putting .fog into the tests folder).   I just checked if a provisioned fog-test-server had preemptibility: true.  I didn't write any automated tests to curl the metadata.  Sorry!  My ruby-fu is too weak to write a test. :-/

jcdang:
Oh yeah.  Another thing to mention.  An instance can't be migratable and preemptible so you need to set onHostMaintenance to TERMINATE

ihmccreery:
Cool.  Is that logic built in?  I don't see it, but I could be missing
something.  What happens if a user tries to create a preemptible VM without
setting TERMINATE?

jcdang:
The logic isn't in but you'll get the error when the api response comes back and a fog error is thrown.

GCE won't let you create the box.

ihmccreery:
Thanks jcdang!

PR #56, opened 2015-07-22
Allowing whitespace when uploading to google cloud

richardaustenwallace:
Currently, fog-google uses `CGI.escape` when sanitizing the names of files to be uploaded to Google Cloud. This causes `my file.txt` to become `my+file.txt` . I copied over the method used in `Rackspace.escape` to allow whitespace characters in filenames.

### DIFF
--- a/lib/fog/google.rb
+++ b/lib/fog/google.rb
@@ -28,6 +28,16 @@ module Fog
     service(:storage,    'Storage')
     service(:sql,        'SQL')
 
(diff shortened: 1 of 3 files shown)


icco:
Seems to be ok according to https://cloud.google.com/storage/docs/bucket-naming

icco:
Thanks!

dambalah:
Thanks! :+1:

PR #64, opened 2015-09-17
Update dns.rb to use v1 rather than v1beta1

canausa:
v1beta1 will deprecated eventually lets move to v1. This should be a no-op.


Temikus:
Note to maintainers: v1beta1 will be deprecated _very soon_ (docs state cutoff on Sept 25th).

mbrukman:
@erjohnso, @icco, @ihmccreery — can one of you please review + commit this ASAP? The v1beta1 API will stop working on 25 Sep 2015. Thanks!

erjohnso:
Thanks for this @canausa!

(And thanks for the heads-up @Temikus and @mbrukman) :)

PR #68, opened 2015-09-27
Update CI for Ruby <1.9.3 deprecation

Temikus:
As per @plribeiro3000 comment in: fog/fog-core#162 we should drop support for the following ruby versions since fog does not support them anymore:

ruby 1.8.7
1.9.2 
ree
jruby-18mode

Looks ok to everyone?

NOTE: Build is currently failing due to https://github.com/fog/fog-core/issues/162


plribeiro3000:
I believe it would be nice to add a notice to the `README`.

icco:
Added readme, merged in #69, I'm extending this to `< 2.0.0` unless I hear otherwise.

plribeiro3000:
:+1:

mmoll:
1.9.3 is in wide usage on LTS supported platforms like Debian/wheezy and Ubuntu/trusty. Forcing this to >= 2.0.0 introduces problems for all people who package software to use Rubies and gems from the system repositories. While I have no clear picture of the real consequences I guess it's making the life of i.e. @theforeman much harder once there's some API change by Google that gets handled only by fog-google versions that are not supporting Rubies <2.0.0 anymore.

plribeiro3000:
@mmoll I agree with you here and thats exactly why this got so much time before it actually happened. What we got to decide is that the ruby community (i.e. other gems) is dropping support of it so is harder and harder to keep it floating around here.

With that being said i must say that we can rollback this if we really need (broken OS and stuff), otherwise we will be just trying to solve a problem at the wrong spot. Ruby `1.9.3` was discontinued but some OS still use it (not saying it is a wrong or something) and the man in the middle (`fog`) suffers a lot trying to keep it up for both sides.

I admit this is not a clean move but it will at least allow us to move more softly.

Happy to keep discussing it,

Paulo

icco:
Just to be clear, Precise is the oldest supported ubuntu and does not have ruby 2. Trusty does support ruby2 : http://packages.ubuntu.com/source/trusty/ruby2.0. On debian Jessie (the current stable) does support ruby2: https://packages.debian.org/jessie/libruby2.1. If you are on an LTS release on debian though, they pretty explicitly say you shouldn't expect new patches beyond ones provided by volunteers to work: https://wiki.debian.org/LTS/FAQ

daviddavis:
What are the advantages to dropping 1.9.3 support? And how much extra work is required in supporting it?

icco:
It should be stated also @mmoll the specific concern is that net-ssh has dropped support of rubies less than 2.0, and since net-ssh is how most commands are sent to compute, it's one of the more important dependencies. Also, everything @plribeiro3000 said.

.travis.yml
```diff
@@ -1,14 +1,10 @@
 matrix:
   include:
-    - rvm: 1.8.7
-      gemfile: gemfiles/Gemfile.2.0.0-
-    - rvm: 1.9.2
-      gemfile: gemfiles/Gemfile.2.0.0-
     - rvm: 1.9.3
       gemfile: gemfiles/Gemfile.2.0.0-
```
plribeiro3000: Oops, you did not remove this guy over here. :sweat_smile:

mmoll:
Sadly I don't have the time to _really_ get into this, but... ;)

Just saw net-ssh/net-ssh#261 and net-ssh/net-ssh#260 and am unsure if all this really helped them, but I'm getting the root problem and for sure it's not the last time the overly short Ruby support cycles are causing churns.

Regarding Ubuntu/trusty, after all the problems we had with just getting a newer Ruby than 1.9.3-p0 on Ubuntu/precise and dealing with the fact that 1.8.7 is still the default while 1.9 was also coming as an option I see the situation on trusty a bit more relaxed with 2.0 but it will also be not a pure no-op.

In general, this is not fog's fault at all but a consequence of the behaviour of quite some developers not taking the situation of people into account that can't/dont want to use rvm/rbenv/self-compiled Rubies on their systems.

davexunit:
This patch has broken compatibility with Ruby < 2.0.0, even when you do not want to use the Google API. The Fog documentation says: "Fog officially supports Ruby 2.1.1, 2.1.0, 2.0.0, 1.9.3, 1.9.2, and 1.8.7" [0], but this patch implicitly changes that to "Fog officially supports Ruby 2.1.1, 2.1.0, 2.0.0."  I have an application that is still using 1.9.3 and upgrading to the latest 1.x Fog release broke things despite the documentation saying the version I used was supported.

Perhaps it would be best to revert this patch and apply it to the 2.x series instead.  WDYT?

[0] https://github.com/fog/fog/blob/master/lib/fog/rackspace/docs/getting_started.md

plribeiro3000:
Yes. We need to review this and fix the whole thing. Unfortunately i'm not having time to spare on this.
It should be easy as changing `fog` to require a version of `fog-google` that does not have this constraint.
This is not something to be fixed here but rather at `fog`.

davexunit:
Thanks for the suggestion.  It works as a short term solution.  Happy hacking!

plribeiro3000:
Sure. We will release a new version of `fog` at the `1.x` series with this fix asap.

geemus:
Sorry for the churn/issues. Due to net-ssh 1.9 is not very tenable and due to nokogiri 1.8 isn't either. I don't want to leave people hanging, but we have also gone out of our way for quite some time to try and accommodate. I don't want to leave people hanging, but I also don't want to not get ssh related updates. Seems like pushing forward is the least evil option for us personally, enforced by perhaps unfortunate choices/timing that lays outside our control.

PR #48, opened 2015-06-04
Introducing rubocop linter

Temikus:
Work towards #15 
I recently introduced linters into another project, so quickly made a foundation for fog-google. 
Those tasks should help you work through the issues gradually and ease CI integration in the future somewhat.

Introducing:
1) Linter tasks in a :lint namespace:
```
λ rake lint
lint               -- run rubocop
lint:auto_correct  -- auto-correct rubocop offenses
```
2) .rubocop_todo.yml file, to work through the issues at your own pace.

What do you think?

Issue #34, opened 2015-05-29
attribute names should be standardized


ihmccreery:
For example, `Server` has the following format for attributes, (this seems to be the preferred format):

```ruby
attribute :can_ip_forward, :aliases => 'canIpForward'
attribute :creation_timestamp, :aliases => 'creationTimestamp'
```

whereas `UrlMap` has the following format:

```ruby
attribute :fingerprint, :aliases => 'fingerprint'
attribute :hostRules, :aliases => 'host_rules'
```

It's worth noting that the format that `UrlMap` provides *does not* allow `:underscored_symbol` notation as it stands now, (symbol keys are not automatically converted to or checked against strings when passing parameters around).

PR #79, opened 2015-10-24
Lint cleanup

icco:
@plribeiro3000 not sure who I should send my code reviews to these days... not sure what the status is over in Google employee land. But I ran `rubocop -a` on the directory for giggles, and got this. 

All of it makes sense, and tests pass.

/cc @ihmccreery @Temikus


icco:
This is, in my mind, progress towards #34.

plribeiro3000:
@icco I think this is pretty straight forward. If you think this is good to go lets merge it. =)

plribeiro3000:
:+1: Maybe we should add `rubocop` to the build. What do you think?

Temikus:
@plribeiro3000 @icco I was thinking of ressurecting #48, since we don't rely on 1.9.3. anymore. Would that work?

plribeiro3000:
@Temikus Ops. Didn't know you had already work towards this. Sorry.
Since this one was merged i believe we can move from here. We just need to make `rubocop` run in Travis as well. 

I'm not sure what you guys think but i'm using @houndci to handle this and i kind of like it. =)
It is a spammer but at least i don't have to go to the build to see whats wrong. =)

Temikus:
@plribeiro3000 Not a problem at all :) With this merged, the only thing left is to add a rake task and add it to travis. I think with a variable `LINTER=TRUE` or something.

I'm all for trying to use hound. The question is whether their free tier is allowed for our use-case.
And, AFAIK it's only a linter, so not exactly replacement for our Travis setup.

icco:
I'm down with @houndci or just adding it to the build or whatever. Thankfully github allows for multiple CIs on the same pull request.

Issue #78, opened 2015-10-18
Fog::DNS::Google::Records example is broken


Temikus:
get method in Fog::DNS::Google::Records supports only positional arguments

`connection.records.get` is structured like so:
```
def get(name, type)
  requires :zone

  records = service.list_resource_record_sets(zone.identity, { :name => name, :type => type }).body['rrsets'] || []
  records.any? ? new(records.first) : nil
rescue Fog::Errors::NotFound
  nil
end
```
, however, our example lists:
```
record = connection.records(zone: zone).get(name: 'tessts.example.org.',type: 'A')
```
, which leads to argument errors:
```
> record = connection.records(zone: zone).get(name: 'tessts.example.org.',type: 'A')
ArgumentError: wrong number of arguments (1 for 2)
```

Should I:
 - Fix the example?  
or
 - Fix the method to accept an options hash? ( Ruby 2.0 doesn't support _required_ named arguments :disappointed: )

/CC @plribeiro3000 @icco

icco:
Please fix the example. Thanks for catching!

Temikus:
Fixed by #81

icco:
<3 ty.

PR #81, opened 2015-10-24
Fixing zones example

Temikus:
Fixing #78


icco:
Technically we're still using the style choice of hashrockets ;)

Temikus:
@icco Oh, good catch! I'll fix it up.

PR #82, opened 2015-10-25
Consistent alias defintions

### LINKED ISSUE #34: attribute names should be standardized

icco:
Fixes #34.

Also goes up and cleans up some rubocop warnings in `models/`

### DIFF
--- a/lib/fog/google/models/compute/url_map.rb
+++ b/lib/fog/google/models/compute/url_map.rb
@@ -21,11 +21,11 @@ module Fog
           requires :name, :default_service
 
           options = {
-            "defaultService" => defaultService,
+            "defaultService" => default_service,
             "description" => description,
             "fingerprint" => fingerprint,
-            "hostRules" => hostRules,
-            "pathMatchers" => pathMatchers,
+            "hostRules" => host_rules,
+            "pathMatchers" => path_matchers,
             "tests" => tests
           }
 
--- a/lib/fog/google/models/compute/url_map.rb
+++ b/lib/fog/google/models/compute/url_map.rb
@@ -18,7 +18,7 @@ module Fog
         attribute :tests, :aliases => "tests"
 
         def save
-          requires :name, :defaultService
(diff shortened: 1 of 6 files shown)


lib/fog/google/models/compute/url_map.rb
```diff
@@ -7,25 +7,25 @@ class UrlMap < Fog::Model
...
           options = {
-            "defaultService" => defaultService,
+            "defaultService" => default_service,
             "description" => description,
             "fingerprint" => fingerprint,
-            "hostRules" => hostRules,
-            "pathMatchers" => pathMatchers,
+            "hostRules" => host_rules,
+            "pathMatchers" => path_matchers,
             "tests" => tests
```
plribeiro3000: This is not related to your changes but ´ve faced this same pattern before and i ended up adding a new feature to `fog-core`, the `as` option like [here](https://github.com/fog/fog-xenserver/blob/master/lib/fog/compute/xen_server/models/server.rb#L29).
Then my `save` method is as simple as [this](https://github.com/fog/fog-xenserver/blob/master/lib/fog/compute/xen_server/models/instance_methods.rb#L22-L28) because the method `all_associations_and_attributes` will return a hash following the correct convetion. 
Hope this helps here too.

icco: Cool, filed https://github.com/fog/fog-google/issues/86

PR #83, opened 2015-10-25
Storage delete_object fix + new storage example

Temikus:
#56 implemented a new escape method for create_object and copy_object, but didn't change the
delete_object method, leading to a situation when we can create a file with whitespace in it, but cannot actually delete it.

Additionally, added a new Google storage usage example.


plribeiro3000:
:+1:

PR #87, opened 2015-11-01
Removing <=1.9 rules from gemspec

Temikus:
We still had Ruby 1.9 version fixup in the gemspec for rubocop and coveralls.
Considering the gemspec now doesn't allow <2.0, I presume we can remove it.


plribeiro3000:
:+1:

PR #90, opened 2015-11-12
make net-ssh require optional

geemus:


geemus:
@Temikus I think this should fix the issue you mentioned, but could definitely use help confirming. Basically I think without this it should (hopefully) not blow up without net/ssh and should work if you explicitly require.

plribeiro3000:
@geemus I think there is only one catch up here. The current mast of `fog-google` has `1.9.3` incompatible code already. This will fix the `net/ssh` issue but won't fix the `1.9.3` issue.

Perhaps this should be agains a legacy branch or tag?

Issue #31, opened 2015-05-03
Implement better examples


Temikus:
Сurrently all fog-google examples are essentially code snippets:
Take https://github.com/fog/fog-google/blob/master/examples/image_all.rb as an example:

```
def test
  connection = Fog::Compute.new({ :provider => "Google" })

  # If this doesn't raise an exception, everything is good.
  connection.images.all
end
```

If someone wants to fix something quick, he needs to first set up a development environment, which potentially deters quick changes to the library and getting started easily.

What I propose is:
1) Briefly describe in the README to get `.fogrc` working, as in:

```
default:
    google_project: "my-awesome-project"
    google_client_email: "some-numbersandsuch@developer.gserviceaccount.com"
    google_json_key_location: "/tmp/gce-key.json"
```

2) Adding development dependencies to a separate group in Gemfile, for example, for latest git versions of fog and fog-core:

```
group :development do
   gem 'fog-core', git: "https://github.com/fog/fog-core"
   gem 'fog', git: "https://github.com/fog/fog"
   gem 'fog-google', path: "."
   gem 'fog-json'
end
```

3) With that set up we just need to add 4 lines to any example:

```
require 'bundler'
Bundler.require(:default, :development)
# Comment this if you don't want to make real requests to GCE
WebMock.disable!
```

And voila! It becomes an actual working script:

```
temikus λ cat image_all.rb
require 'bundler'
Bundler.require(:default, :development)
# Comment this if you don't want to make real requests to GCE
WebMock.disable!

def test
  connection = Fog::Compute.new({ :provider => "Google" })

  # If this doesn't raise an exception, everything is good.
  connection.images.all
end

images = test
pp test

temikus λ ruby image_all.rb >> log
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.6-compliant syntax, but you are running 2.1.5.
[  <Fog::Compute::Google::Image
    name="centos-6-v20131120",
    id="11748647391859510935",
    kind="compute#image",
    archive_size_bytes="269993565",
    creation_timestamp="2013-11-25T15:13:50.611-08:00",
    deprecated={"state"=>"DEPRECATED", "replacement"=>"https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/centos-6-v20150226"},
    description="SCSI-enabled CentOS 6 built on 2013-11-20",
    disk_size_gb="10",
[...]

PR #94, opened 2015-12-06
Cleaning up examples

Temikus:
Take 1 on #31, some are still left as they are
- Made the examples easily accessible.
- Removed old and redundant ones.
- Updated resources, images, etc.

/CC @plribeiro3000


examples/eric-fail.rb
```diff
@@ -1,32 +0,0 @@
-def test
-  connection = Fog::Compute.new(:provider => "Google")
```
icco: lol, I guess we can stop making fun of @erjohnso in the fog-google examples :p

erjohnso:
@icco - there is no legitimate reason I can think of to stop making fun of me.

examples/eric-fail.rb
plribeiro3000: lol. did not know about this file at all. :sweat_smile:

icco: I can't remember the original bug, but there was some weird way that @erjohnso was using Fog in another library, which caused this script to blow up. It was interesting at the time, now it's just funny :p

Issue #50, opened 2015-06-08
Shindo should be replaced with minitest


ihmccreery:
Per fog/fog#1266 and fog/fog#2630, Minitest is the new testing framework to be used by fog projects.
- [x] #51 is open to cover `compute` in integration tests.
- [ ] Nothing is unit-tested with `minitest`; it should be.  Opening #49.
- [ ] Furthermore, it's not clear whether or not we can run these integration tests in Travis CI, (see #19).  Regardless, unit tests should be run in Travis.
- [ ] #51 only covers `compute`.  `dns`, `monitoring`, `sql`, and `storage` all remain tested in under shindo, and should be moved over to `minitest`.
- [x] The Rakefile is kind of a mess, and should eventually be cleaned up.

icco:
Sub issues have mostly been closed.

PR #96, opened 2015-12-06
New JSON Storage API

icco:
This pull request is not ready at all, but creating it so we can finally make progress on #38.

Goal for this before it's merged: Auth and get files with tests.

/cc @deanputney, @Temikus and @plribeiro3000 for FYI. Feel free to leave comments and what not.


lib/fog/google/models/storage_json/file.rb
```diff
@@ -0,0 +1,146 @@
...
+        end
+
+        # TODO: Verify
+        remove_method :metadata=
+        def metadata=(new_metadata)
+          merge_attributes(new_metadata)
+        end
+
+        # TODO: Verify
+        remove_method :owner=
```
plribeiro3000: Why not just overwrite it?

icco: Direct copy from the storage folder. The storage code is from before my time, I think I'm gonna have to nuke most of this, but I want to keep the API the same except for Auth.

plribeiro3000: Got it.

2015-12-07 12:01 GMT-02:00 Nat Welch <notifications@github.com>:

> In lib/fog/google/models/storage_json/file.rb
> <https://github.com/fog/fog-google/pull/96#discussion_r46822952>:
>
> > +        end
> > +
> > +        # TODO: Verify
> > +        remove_method :metadata
> > +        def metadata
> > +          attributes.reject { |key, _value| !(key.to_s =~ /^x-goog-meta-/) }
> > +        end
> > +
> > +        # TODO: Verify
> > +        remove_method :metadata=
> > +        def metadata=(new_metadata)
> > +          merge_attributes(new_metadata)
> > +        end
> > +
> > +        # TODO: Verify
> > +        remove_method :owner=
>
> Direct copy from the storage folder. The storage code is from before my
> time, I think I'm gonna have to nuke most of this, but I want to keep the
> API the same except for Auth.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fog/fog-google/pull/96/files#r46822952>.
>



-- 
-----------------------------------------------------
Paulo Henrique Lopes Ribeiro
Senior Ruby Engineer

plribeiro3000:
So whats the thing with this `StorageJSON`?

icco:
@plribeiro3000 I've updated the readme in this PR, see if that helps explain WTF is up and why this is needed.

icco:
P.S. @erjohnso, @deanputney is going to contribute some of his first fog code (and got me to finally start working on this issue) and lives in Seattle, you guys should meet!

Temikus:
@icco See some minor comments - otherwise - looks cool so far. It would be awesome to have proper json API support. Thanks for starting this! :+1: :fireworks:

plribeiro3000:
@icco Wow! Thats a great README. =)

Thank you for taking the time to do so. I was expecting something on those lines but was not sure. :+1:

icco:
One thing agreed upon by @deanputney and I, this version will not support resumable uploads.

README.md
```diff
@@ -2,42 +2,46 @@
...
 
 ## Storage
 
-Google Cloud Storage originally was very similar to Amazon's S3. Because of this, Fog implements the [XML GCS API](https://developers.google.com/storage/docs/xml-api-overview). We eventually want to move to the new [JSON API](https://developers.google.com/storage/docs/json_api/), once it has similar performance characteristics to the XML API. If this migration interests you, send us a pull request!
+There are two ways to access [Google Cloud Storage](https://cloud.google.com/storage/). The old S3 API and the new JSON API.
+
+ * The [XML API](https://developers.google.com/storage/docs/xml-api-overview) is almost identical to S3. It is accessed through `Fog::Storage::Google`.
+ * The new [JSON API](https://developers.google.com/storage/docs/json_api/) is faster and uses auth similarly to the rest of the Google Cloud APIs. It is accessed through `Fog::Google::StorageJSON`.
+
+Eventually the XML API will be deprecated in favor of the JSON API, but we have no idea on the timeline for that.
```
erjohnso: Let's whack line 16. Last time I checked, GCS team decided the XML API was worth keeping for the foreseeable future.

icco:
Open work item:

"Write a test (and relevant working code) for the paperclip use case, where a user just shovels in auth to Fog::Storage.new. Those options should allow for service accounts, old xml auth stuff or new JSON creds."

My theory is we need to modify the constructor of Fog::Storage::Google to return Fog::Google::StorageJson when json creds are provided and Fog::Storage::Google when xml creds are provided

deanputney:
Just want to jump in here to say hi, since I've been working away on this but haven't gotten in this conversation much or introduced myself. Also, as of today the JSON API is functioning with Paperclip, albeit in a limited manner. There's still a bit of work to be done around ACLs and permissions, but manipulating files and directories is working great.

icco:
:wave: Great work so far! Glad to hear our solution for Paperclip worked @deanputney!

icco:
cc @Temikus and @plribeiro3000, this is basically ready to go. I'm gonna do one more read through tonight, and try and clean up some of the @houndci errors, but other than that, I think it should be merged. @deanputney killed it!

plribeiro3000:
nice @Temikus! This seems good to go. Every time i got an @houndci email i did a look because that means more work was being pushed.

I will try to take another look but AFAICT this is good.

README.md
```diff
@@ -2,42 +2,44 @@
...
+Fog implements [v2beta1](https://cloud.google.com/monitoring/v2beta2/) of the Google Cloud Monitoring API. This is a currently deprecated version of the API. Pull requests for updates would be greatly appreciated.
 
 ## Installation
 
-Add this line to your application's Gemfile:
+Add the following two lines to your application's `Gemfile`:
 
 ```ruby
 gem 'fog-google'
+gem 'google-api-client'
```
Temikus: I would set it to something like this, otherwise we may get issues with people using either a very old untested client or the incompatible 0.9 version:
gem 'google-api-client', '< 0.9', '>= 0.6.2'

examples/storage_json.rb
```diff
@@ -0,0 +1,37 @@
+# All examples presume that you have a ~/.fog credentials file set up.
+# More info on it can be found here: http://fog.io/about/getting_started.html
+
+require "bundler"
+Bundler.require(:default, :development)
+# Uncomment this if you want to make real requests to GCE (you _will_ be billed!)
+WebMock.disable!
```
Temikus: Comment this down so it doesn't produce real requests from the start.

lib/fog/google/models/storage_json/directories.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+
+        def get(key, options = {})
+          remap_attributes(options,             :delimiter  => "delimiter",
+                                                :marker     => "marker",
+                                                :max_keys   => "max-keys",
+                                                :prefix     => "prefix")
+          data = service.get_bucket(key, options).body
+          directory = new(:key => data["name"])
+          # options = {}
```
Temikus: Is this commented block relevant?

lib/fog/google/models/storage_json/file.rb
```diff
@@ -0,0 +1,154 @@
...
+        @valid_predefined_acls = %w(private projectPrivate bucketOwnerFullControl bucketOwnerRead authenticatedRead publicRead)
+
+        def predefined_acl=(new_acl)
+          unless @valid_predefined_acls.include?(new_acl)
+            raise ArgumentError.new("acl must be one of [#{@valid_predefined_acls.join(', ')}]")
+          end
+          @predefined_acl = new_acl
+        end
+
+        # def acl=(new_acl)
```
Temikus: Is this a TODO for object ACL's? If so - can you mark it as such?

lib/fog/google/models/storage_json/file.rb
```diff
@@ -0,0 +1,154 @@
...
+        end
+
+        # TODO: Verify
+        def body=(new_body)
+          attributes[:body] = new_body
+        end
+
+        attr_reader :directory
+
+        # TODO: Verify
```
Temikus: + all other "Verify" comments. Has this been verified? What should be verified?

lib/fog/google/models/storage_json/files.rb
```diff
@@ -0,0 +1,93 @@
...
+  module Storage
+    class GoogleJSON
+      class Files < Fog::Collection
+        extend Fog::Deprecation
+        deprecate :get_url, :get_https_url
+
+        attribute :common_prefixes, :aliases => "CommonPrefixes"
+        attribute :delimiter,       :aliases => "Delimiter"
+        attribute :directory
+        # attribute :is_truncated,    :aliases => "IsTruncated"
```
Temikus: I'm not sure, but seems like it is from the Amazon-compatible API. If so - maybe it should be removed? If not - at least put a TODO

lib/fog/google/models/storage_json/file.rb
```diff
@@ -0,0 +1,154 @@
...
+          target_directory = service.directories.new(:key => target_directory_key)
+          target_directory.files.get(target_file_key)
+        end
+
+        # TODO: Verify
+        def destroy
+          requires :directory, :key
+          begin
+            service.delete_object(directory.key, key)
+          rescue Excon::Errors::NotFound
```
Temikus: Empty rescue block.

lib/fog/google/models/storage_json/files.rb
```diff
@@ -0,0 +1,93 @@
...
+
+            self
+          end
+        end
+
+        def get(key, options = {}, &block)
+          requires :directory
+          data = service.get_object(directory.key, key, options, &block)
+          file_data = {}
+          data.headers.each do |key, value|
```
Temikus: 'key' is shadowing the outer parameter of 'get'. Suggest replacing key, value with |k,v|

lib/fog/google/requests/storage_json/put_object.rb
```diff
@@ -0,0 +1,95 @@
...
+            object = {
+              :body             => data[:body],
+              "Content-Type"    => options["Content-Type"] || data[:headers]["Content-Type"],
+              "ETag"            => Fog::Google::Mock.etag,
+              "Key"             => object_name,
+              "Last-Modified"   => Fog::Time.now.to_date_header,
+              "Content-Length"  => options["Content-Length"] || data[:headers]["Content-Length"]
+            }
+
+            for key, value in options
```
Temikus: +1 (I know it's from the old module, but still)

test/integration/storage/test_objects.rb
```diff
@@ -0,0 +1,114 @@
+require "helpers/integration_test_helper"
+
+class TestObjects < FogIntegrationTest
+  def setup
+    # Uncomment this if you want to make real requests to GCE (you _will_ be billed!)
```
Temikus: Unnecessary comment in integration tests, ditto for the one in buckets and objects.

lib/fog/google/storage_json.rb
```diff
@@ -0,0 +1,146 @@
...
+          query << "Expires=#{params[:headers]['Date']}"
+          "#{params[:host]}/#{params[:path]}?#{query.join('&')}"
+        end
+      end
+
+      class Mock
+        include Utils
+        include Fog::Google::Shared
+
+        def initialize(options)
```
Temikus: Ditto required argument vs options hash + `.fog` file usage.

lib/fog/google/storage_json.rb
```diff
@@ -0,0 +1,146 @@
...
+      end
+
+      class Real
+        include Utils
+        include Fog::Google::Shared
+
+        attr_accessor :client
+        attr_reader :storage_json
+
+        def initialize(options)
```
Temikus: Ditto required argument vs options hash + .fog file usage.

lib/fog/google/storage.rb
```diff
@@ -1,292 +1,14 @@
...
-          params = request_params(params)
-          scheme = params.delete(:scheme)
-          host   = params.delete(:host)
-          port   = params.delete(:port)
-
-          params[:headers]["Date"] = Fog::Time.now.to_date_header
-          params[:headers]["Authorization"] = "GOOG1 #{@google_storage_access_key_id}:#{signature(params)}"
-
-          connection(scheme, host, port).request(params, &block)
+      def self.new(options)
```
Temikus: @icco @deanputney - This now breaks scenarios when Fog::Storage::Google.new is invoked without arguments, for example when `~/.fog` file is used.
Is this by design?

This happens because the old storage connection was invoked with an options hash , while now it takes a required argument.

For repro just run the storage example:
```
ruby storage_json.rb                                                                                                                                          (1)
/Users/temikus/Code/fog-google/lib/fog/google/storage.rb:7:in `new': wrong number of arguments (0 for 1) (ArgumentError)
	from storage_json.rb:14:in `test'
	from storage_json.rb:37:in `<main>'"
```

lib/fog/google/storage.rb
```diff
@@ -1,292 +1,14 @@
...
+        if options.keys.join(" ").include? "key"
```
Temikus: Wouldn't this trigger on HMAC key options as well?

```
   google_storage_access_key_id: "GOOGAVMVXSY2VMMPNSTB"
   google_storage_secret_access_key: "S1zcqqnNmfeEu1IsQwvOJB9ekrqHXxvCF0tHA0Jz"
```

Temikus:
@plribeiro3000 All credit should go to @icco and @deanputney ;)

@icco @deanputney - Please see my comments, especially the ones about required options and the StorageXML vs StorageJSON fork.

Otherwise - looks good. Thank you so much @deanputney and @icco! :fireworks: :fireworks: :fireworks:

examples/storage_json.rb
```diff
@@ -0,0 +1,37 @@
...
+
+# This specific example needs google_storage_access_key_id: and google_storage_secret_access_key to be set in ~/.fog
+# One can request those keys via Google Developers console in:
+# Storage -> Storage -> Settings -> "Interoperability" tab -> "Create a new key"
```
Temikus: This comment on how to get HMAC keys is irrelevant for the JSON example.

lib/fog/google/storage.rb
deanputney: This is a misunderstanding on how Fog::Storage::Google.new can be called. My previous version of this worked with the `~/.fog` credentials but not with the parameters, I guess I need to check both of these options here.

deanputney:
Excellent, thanks for the review. Much needed and appreciated. I'll dig through these items here and get some fixes up soon, we're very excited to get this in place!

lib/fog/google/models/storage_json/directories.rb
deanputney: I'm not really sure. This is one of the places where the JSON API diverges from the XML one and I'm a little confused about what the expected behavior is.

First: none of those keys or anything similar exist in the [Bucket resource](https://cloud.google.com/storage/docs/json_api/v1/buckets#resource) that get_bucket returns. So that loop can be ditched.

Second, however, I'm not sure what the significance of the `directory.files.load(data["contents"])` is. `contents` isn't a valid key in the Bucket resource either, so I doubt this would help us in any way.

It _seems_ like we can ditch this whole block from that. I left it in because I wasn't entirely sure.

lib/fog/google/models/storage_json/file.rb
deanputney: This is unmodified from the [original storage definition](https://github.com/icco/fog-google/blob/new_storage_api/lib/fog/google/models/storage/file.rb#L47). Should I log the error in both places? Something like:

```ruby
rescue Excon::Errors::NotFound => e
    logger.error e
    logger.error e.backtrace
end
```

lib/fog/google/storage.rb
deanputney: Yeah, perhaps the better case is to check on `"client_email"`. Any better suggestions than that?

lib/fog/google/storage.rb
deanputney: Added a fix for this and the options defaulting to `{}`. Coming soon.

test/integration/storage/test_objects.rb
deanputney: What makes that an unnecessary comment? Shouldn't people be aware of that?

lib/fog/google/models/storage_json/directories.rb
Temikus: Looking into blame - this looks like a blind copy from AWS provider:
https://github.com/fog/fog/blame/90e96f36e6f70a12f0c6794a149231c337293887/lib/fog/google/models/storage/directories.rb
Relevant commit: https://github.com/fog/fog/blame/90e96f36e6f70a12f0c6794a149231c337293887/lib/fog/google/models/storage/directories.rb

So if it's not used - I think we can leave it out. @icco @plribeiro3000 - what do you guys think?

lib/fog/google/models/storage_json/file.rb
Temikus: Looks like this was forklifted from AWS as well. Have you checked - does it actually trigger Excon::Errors::NotFound if an object doesn't exist?

If so - I think if the object is not found, the deletion has failed, so we should return false, same as with directories. What do you think?

test/integration/storage/test_objects.rb
Temikus: Ok, so, my understanding is people will not generally want to go and edit tests to run them.
AFAIK - there's no mocking in place for the new set of integration tests, so this will fail anyways.

IMO as soon as they are actually functional, the warning/config option should be on a very high level. For example, when you run `rake test` with no arguments it throws a warning or asks you whether you want to continue with an option for mocks somewhere in top layer test config. + having a default yes option for CI, like `rake test:bill_me_money` or something. What do you think?

lib/fog/google/storage.rb
Temikus: This will need to be tested. Because `~/.fog` credentials file may contain both `google_storage_access_key_id:` and `client_email` since we had no options for JSON storage access before.

The main idea is - we don't want to break people's code if they've been using the old API and just upgraded with their existing config.

lib/fog/google/models/storage_json/directories.rb
plribeiro3000: Dunno. If is not used or does not make sense we can rip it out but i'm not that into the platform to say. :sweat:

deanputney: It might be related to how a `directory` object gets its associated `files`, but I'm not really sure. That might be handled by `list_objects`. In which case, we should ditch it.

So far things seem to be working without it. So maybe let's ditch it and revisit if needed.

lib/fog/google/models/storage_json/file.rb
deanputney: Makes sense to me. I'll update this.

lib/fog/google/storage.rb
deanputney: Yeah, I'm hitting some odd edge cases here and trying to work through them in my PR to `icco/fog-google`. The precise logic for this is important, as you point out, and there are a few cases that throw it off. My current working approximation is:

```ruby
if options.keys.include? :google_storage_access_key_id or
  (Fog.credentials and Fog.credentials.keys.include? :google_storage_access_key_id)
  Fog::Storage::GoogleXML.new(options)
else
  Fog::Storage::GoogleJSON.new(options)
end
```

This handles these cases:

* `options` includes `google_storage_access_key_id` -> use `GoogleXML`
* `options` doesn't include `google_storage_access_key_id`, `Fog.credentials` does -> use `GoogleXML`
* Neither `options` nor `Fog.credentials` include `google_storage_access_key_id` -> use `GoogleJSON`

There's an important failure when `Fog.credentials` is not set, though. Calling `Fog.credentials` without a credential file in place throws an error that stops the whole process. 

I think the logic is better– if that interoperability key is present we should use the XML every time. What would be a good way to handle `Fog.credentials` here though?

examples/storage_json.rb
deanputney: :white_check_mark:

README.md
deanputney: :white_check_mark:

lib/fog/google/models/storage_json/directories.rb
deanputney: :white_check_mark:

lib/fog/google/models/storage_json/file.rb
deanputney: :white_check_mark:

lib/fog/google/models/storage_json/files.rb
deanputney: :white_check_mark:

lib/fog/google/models/storage_json/file.rb
deanputney: :white_check_mark:

lib/fog/google/models/storage_json/files.rb
deanputney: :white_check_mark:

lib/fog/google/storage_json.rb
deanputney: :white_check_mark:

lib/fog/google/storage_json.rb
deanputney: :white_check_mark:

test/integration/storage/test_objects.rb
deanputney: I'd defer to you on this. If there's somewhere higher up to put this warning it sounds like the right thing to do to me. These tests could use some work, and I'm not familiar enough with the stack for `rake test` to suggest a better place to include this warning (although I'm sure there is one).

Temikus: Thanks! Don't worry about it for now, we'll address it in #50

lib/fog/google/models/storage_json/file.rb
```diff
@@ -0,0 +1,154 @@
...
+            Fog::Logger.deprecation("options param is deprecated, use acl= instead [light_black](#{caller.first})[/]")
+          end
+          options["contentType"] = content_type if content_type
+          options["predefinedAcl"] ||= @predefined_acl if @predefined_acl # predefinedAcl may need to be in parameters
+          options["acl"] ||= @acl if @acl # Not sure if you can provide both acl and predefinedAcl
+          options["cacheControl"] = cache_control if cache_control
+          options["contentDisposition"] = content_disposition if content_disposition
+          options["contentEncoding"] = content_encoding if content_encoding
+          options["md5Hash"] = content_md5 if content_md5
+          options["crc32c"] = crc32c if crc32c
```
deanputney: Here's a question: `md5Hash` and `crc32c` are defined in the StorageXML code as they are here. However, changing the body of a file and submitting it returns an error now saying that these values don't match, obviously because they haven't been recomputed after the file contents changes. The same save function succeeds if I leave these out.

I can't find anything that recomputes these hashes in the previous software. Were they just being left out before? Should we bother computing them beforehand?

lib/fog/google/models/storage_json/file.rb
deanputney: I've removed all these except one, which I've now made a TODO to implement.

lib/fog/google/requests/storage_json/get_bucket.rb
```diff
@@ -0,0 +1,88 @@
...
+        end
+      end
+
+      class Mock
+        def get_bucket(bucket_name, options = {})
+          raise ArgumentError.new("bucket_name is required") unless bucket_name
+          response = Excon::Response.new
+          name = /(\w+\.?)*/.match(bucket_name)
+          if bucket_name == name.to_s
+            if bucket = data[:buckets][bucket_name]
```
Temikus: I know this is taken from the old API but no one should '=' in conditional without at least wrapping it.
Since this Mock looks a bit broken anyway, I would just wrap it.

Example from styleguide:
```
# bad (+ a warning)
if v = array.grep(/foo/)
  do_something(v)
  ...
end

# good (MRI would still complain though)
if (v = array.grep(/foo/))
  do_something(v)
  ...
end

# good
v = array.grep(/foo/)
if v
  do_something(v)
  ...
end
```

lib/fog/google/requests/storage_json/put_object.rb
```diff
@@ -0,0 +1,118 @@
...
+          }
+
+          body_object = {
+            :contentType => mime_type,
+            :contentEncoding => options["contentEncoding"],
+          }
+          body_object.merge! options
+
+          acl = []
+          case options["predefinedAcl"]
```
Temikus: Case block with no else.

lib/fog/google/requests/storage_json/get_object_torrent.rb
```diff
@@ -0,0 +1,13 @@
+module Fog
+  module Storage
+    class GoogleJSON
+      class Real
+        # Get torrent for an Google Storage object
+        # * Deprecated
+        def get_object_torrent(_bucket_name, _object_name)
```
Temikus: Wouldn't this fail with ArgumentMissing?

lib/fog/google/requests/storage_json/get_object_url.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Storage
+    class GoogleJSON
+      class Real
+        # Get an expiring object url from Google Storage
+        # Deprecated, redirects to get_object_https_url.rb
+
+        def get_object_url(bucket_name, object_name, expires)
```
Temikus: Wouldn't this fail with ArgumentMissing?

lib/fog/google/models/storage_json/file.rb
Temikus: Yeah, I'm not sure about MD5. Before it was supplied as the usual Content-MD5 field. If you've read [this](https://cloud.google.com/storage/docs/hashes-etags#_JSONAPI), you know as much as I do. Maybe @icco or @plribeiro3000 can chime in?

Temikus:
@deanputney Thanks again for the massive amount of work! :+1: 
Checked out the changes after feedback and gave some additional minor comments. 

My use-case with a cred file is now working correctly, awesome \o/ 
Only thing is if your Fog credentials file has XML keys in it, it seems it will always default to the old API but I don't see this as a serious issue.

So, to sum it up: just address the linter errors and, if possible, do some squashing beforehand (just the linter fixes, etc, so the blame doesn't look excessively noisy) and this looks good to go!

And again, huge thanks to @deanputney and @icco!

erjohnso:
Thanks @deanputney, @icco, and @Temikus!  You guys rock!

I'd love to send out a little Google Cloud swag gift basket to you to show my appreciation. If you're interested, drop me a note with your t-shirt size and postal address to erjohnso at google dot com.

lib/fog/google/models/storage_json/file.rb
plribeiro3000: @Temikus I'm afraid i don't have enough knowledge to help regarding this.
@icco is the man you are looking for. =)

icco:
I'm going to merge this tomorrow @deanputney unless you say anything to stop me.

erjohnso:
Thanks a ton for this @icco, @deanputney and for all the reviews @Temikus and @plribeiro3000!

plribeiro3000:
@erjohnso I actually didn't do much here. It was more on the other guys on this one.

PR #101, opened 2016-02-17
Update fog-google.gemspec

jjasghar:
- `google-api-client` is required for fog-google to work, otherwise a warning
```
[fog][WARNING] Please install the google-api-client gem before using this provider
```
get thrown. This PR moves it from a development dependency to a dependency so it happens in the background.

### DIFF
--- a/fog-google.gemspec
+++ b/fog-google.gemspec
@@ -25,9 +25,10 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fog-core"
   spec.add_dependency "fog-json"
   spec.add_dependency "fog-xml"
+  spec.add_dependency "google-api-client", "< 0.9", ">= 0.6.2"
+
 
   # TODO: Upgrade to 0.9, which is not compatible.
-  spec.add_development_dependency "google-api-client", "< 0.9", ">= 0.6.2"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "shindo"
   spec.add_development_dependency "minitest"


jjasghar:
Pinging  @icco, @Temikus and @plribeiro3000

plribeiro3000:
@jjasghar I believe the main reason is that we don't want to add provider specific dependencies in a way that people that is using `joyent` wont have to install a dependency for something he is not using.

I think that after we finish extracting the providers this might be a thing to reconsider.

Is that blocking your work or something like that?

jjasghar:
@plribeiro3000 actually blocking us over on chef-provisioning-fog. https://github.com/chef/chef-provisioning-fog/pull/162. I'd prefer this to come from the upstream because if you are using `fog-google` fog will throw that warning.

plribeiro3000:
Could you guys push that dependency to the final user? Is that possible?

jjasghar:
Yes, It's possible for us to put it in our gem as a dependency. But it's odd that it isnt taken care of from yall.

plribeiro3000:
@jjasghar Agreed with you here. But at the moment this is a pain point for us to take care.
Can we have this as a workaround for now? Do you think it is a good short time solution?

jjasghar:
Yeah, i'm putting in the PR to do the workaround. I'll comment it here in a bit, but it would be nice to revisit down the line.

plribeiro3000:
@jjasghar Count on me to revisit it. I'm pretty sure this is in @geemus plans as well. :+1:

jjasghar:
https://github.com/chef/chef-provisioning-fog/pull/165 is the work around.

plribeiro3000:
Nice. Can we close this in the meantime?
If you want we could have a issue for this in the main repository because `fog-google` is not the only provider with such problem.

jjasghar:
Yep I'll close :dash:

plribeiro3000:
Thanks!

icco:
Just for historical data here: We don't include the google-api gem as a dependency because that would require all fog users to require the gem, even if they weren't using google. One of the reasons for fragmenting the fog code base is to eventually solve this (so people can just require fog-google), at which  time we'll add it as a dep.

jjasghar:
@icco Ah that makes some sense. Thanks for the history.

icco:
Your welcome. We'd love to add it as a dep, because then we could fix the problem that 0.9 adds breaking changes... And also get customers off of 1.9...

PR #102, opened 2016-02-27
Move Compute files to the expected path

plribeiro3000:
cc/ @Temikus 

Ongoing process to make `fog-google` follow the standards and avoid performance issues when loading the entire `fog` ecosystem as in fog/fog#3430.


plribeiro3000:
Tests are failing because travis fails to install ruby `2.3`. :neutral_face:

Temikus:
Looked it over did some manual testing too to be sure, LGTM :+1: 

@plribeiro3000 Thank you so much \o/ :+1:

PR #103, opened 2016-02-27
Move DNS files to the expected path

plribeiro3000:
cc/ @Temikus 

Ongoing process to make fog-google follow the standards and avoid performance issues when loading the entire fog ecosystem as in fog/fog#3430

Consider merging after #102 .


Temikus:
LGTM :+1: 
Manually tested just in case.

Thanks @plribeiro3000 !

PR #106, opened 2016-03-12
Add support for partial content to support range header.

tronathan:
The gem seems to support the http `Range:` header, however it doesn't accept the `206` response from google, which is required. (206 = Partial Content)

The following PR adds basic support for accepting `206 Partial Content`, allowing the `Range` header to be used to get objects from GCS.

### DIFF
--- a/lib/fog/google/requests/storage/get_object.rb
+++ b/lib/fog/google/requests/storage/get_object.rb
@@ -42,7 +42,7 @@ module Fog
 
           params[:response_block] = Proc.new if block_given?
 
-          request(params.merge!(:expects        => 200,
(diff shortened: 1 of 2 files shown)


icco:
Hilarious, [the docs](https://cloud.google.com/storage/docs/reference-headers#range) say nothing about a different header coming back. Thanks for this @tronathan. 

Please rebase or merge master and I'll merge this.

cc @erjohnso on the documentation bug.

erjohnso:
Thanks for the fix @tronathan and thanks for flagging @icco.  Put in a cl for the docs fix.

PR #108, opened 2016-03-15
Update monitoring for v2beta2

erjohnso:
Only real update was changing the API version to `v2beta2` and updating the name of the scope, otherwise, no real changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,9 @@ Fog implements [v1](https://cloud.google.com/dns/api/v1/) of the Google Cloud DN
 
 ## Monitoring
 
-Fog implements [v2beta1](https://cloud.google.com/monitoring/v2beta2/) of the Google Cloud Monitoring API. This is a currently deprecated version of the API. Pull requests for updates would be greatly appreciated.
+Fog implements [v2beta2](https://cloud.google.com/monitoring/v2beta2/) of the Google Cloud Monitoring API. As of 2016-03-15, we believe Fog for Google Cloud Monitoring is feature complete. We are always looking for people to improve our code and test coverage, so please [file issues](https://github.com/fog/fog-google/issues) for any anomalies you see or features you would like.
+
+ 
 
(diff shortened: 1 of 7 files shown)


plribeiro3000:
Hey @erjohnso. Do you need all those `require 'fog'` ?
I guess is already loaded at that point, no?

erjohnso:
Hi @plribeiro3000, likely not, but I'll be darned if I could figure out how to run those examples. They're also not very consistent, so I wasn't sure the right way to go there (even after sifting through #50).  Happy to learn the RightWay(tm) to update those examples so I can manually test.

Temikus:
@erjohnso - the most convenient way would be with bundler headers. See storage.rb as an example:

# All examples presume that you have a ~/.fog credentials file set up.
# More info on it can be found here: http://fog.io/about/getting_started.html

require "bundler"
Bundler.require(:default, :development)
# Uncomment this if you want to make real requests to GCE (you _will_ be billed!)
WebMock.disable!

Temikus:
P.S. If those are still confusing / not easy to work with - please, let me know what issues you had so I can address them, it's on my todo list to adjust all of them to the new format as in #94

erjohnso:
@Temikus, I had tried adding those lines (like what exist in the compute examples), but what is the command to execute the example(s)?  I had assumed it would be,

```sh
bundle exec ruby examples/monitoring/metrics_descriptors.rb
```

But that throws errors. I'm using rbenv with 2.2.4 if that makes a difference.

erjohnso:
Resolved the issue and @Temikus let me know the *only* way the examples work is if you uncomment the Webmock line.  We chatted a bit and he has some good ideas on making some improvements for the examples.

In the meantime, I think this is ready to go if you folks agree.  Looks like the travis failure is unrelated.

Temikus:
LGTM :+1:
Thanks @erjohnso!

PR #110, opened 2016-03-17
Add list_images method to return current, non-deprecated images

erjohnso:
Currently, there are roughly 420 images across all GCE "official" image projects, but only about 20 that are not deprecated across the same set of projects.  I think most users will only want to see the "current" non-deprecated list.


icco:
:+1:

erjohnso:
Thanks @icco!

PR #111, opened 2016-03-20
Extract inner classes to its own files

plribeiro3000:
This refactor have 3 main reasons:

* Improve Readability
* Make it easier to find inner classes
* Decrease Load time and Number of objects in memory during start up

cc/ @Temikus


lib/fog/storage/google_xml/real.rb
```diff
@@ -0,0 +1,109 @@
...
+          @path_style = options[:path_style] || false
+        end
+
+        def reload
+          @connection.reset if @connection
+        end
+
+        def signature(params)
+          string_to_sign =
+            <<-DATA
```
Temikus: @plribeiro3000 
This now causes XML requests to fail because of whitespace in signature.
I would suggest either to rework it so it doesn't use DATA block or add `# rubocop:disable IndentAssignment` so it doesn't get picked up by Hound.

plribeiro3000: Sure. Totally my bad by missing this.

icco: Couldn't we also just fix by striping whitespace?

plribeiro3000:
@Temikus @icco Fixed. =)

icco:
This looks good to me. @Temikus, I'll leave the final decision and merge up to you.

Issue #71, opened 2015-09-29
project-private acl not allowed


gcolson:
Hi,

It would be great to add the project-private predefined acl (https://cloud.google.com/storage/docs/access-control#predefined-project-private) to the list of valid acl here : https://github.com/fog/fog-google/blob/master/lib/fog/google/models/storage/directory.rb#L13

icco:
Sure that's pretty straight forward.

Temikus:
Fixed by #112

PR #112, opened 2016-03-21
Update list of ACLs and change default private

jayschab:
Add all predefined ACLs currently listed on the Google docs:
https://cloud.google.com/storage/docs/access-control.

Also changed the `public=` method to choose "project-private" in
the false case. (This is the default bucket ACL policy).

I'm not sure there is a best `public=false` choice, but from what I could tell it would involve some updates to `fog` and potentially `paperclip` to allow passing through the desired ACL. 

These addresses issue #71.

### DIFF
--- a/lib/fog/google/version.rb
+++ b/lib/fog/google/version.rb
@@ -1,5 +1,5 @@
 module Fog
   module Google
-    VERSION = "0.1.1"
+    VERSION = "0.1.0"
   end
 end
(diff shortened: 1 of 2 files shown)


lib/fog/google/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module Google
-    VERSION = "0.1.1"
+    VERSION = "0.1.0"
```
icco: Could you please revert this file?

icco:
If you revert the change to the version file, this should merge cleanly. Once you do that @jayschab, I'd love to merge this. The other hound errors are small, you can ignore or fix.

PR #118, opened 2016-04-19
Add initial support for Google Pubsub

selmanj:
This change adds some (not all) support for Google Pubsub (https://cloud.google.com/pubsub/overview)

The basics are there (topic creation/deletion/publishing, subscription creation/deleting/pulling, message acknowledging). I also added a very simple mock implementation as well as an example.

### DIFF
--- a/lib/fog/google/models/pubsub/received_message.rb
+++ b/lib/fog/google/models/pubsub/received_message.rb
@@ -1,0 +1,40 @@
+require "fog/core/model"
+
+module Fog
+  module Google
+    class Pubsub
+      # Represents a ReceivedMessage retrieved from a Google Pubsub
+      # subscription. Note that ReceivedMessages are immutable.
+      #
+      # @see https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/pull#ReceivedMessage
+      class ReceivedMessage < Fog::Model
+        identity :ack_id, :aliases => "ackId"
+
+        attribute :message
+
+        def initialize(new_attributes = {})
+          # Here we secretly store the subscription name we were received on
+          # in order to support #acknowledge
+          attributes = new_attributes.clone
+          @subscription_name = attributes.delete(:subscription_name)
+          super(attributes)
+        end
+
+        # Acknowledges a message.
+        #
+        # @see https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/acknowledge
+        def acknowledge
+          requires :ack_id
+
+          service.acknowledge_subscription(@subscription_name, [ack_id])
+          nil
+        end
+
+        def reload
+          # Message is immutable - do nothing
+          Fog::Logger.warning("#reload called on immutable ReceivedMessage")
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/google/models/pubsub/subscription.rb
+++ b/lib/fog/google/models/pubsub/subscription.rb
@@ -1,0 +1,86 @@
+module Fog
+  module Google
+    class Pubsub
+      # Represents a Pubsub subscription resource
+      #
+      # @see https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions
+      class Subscription < Fog::Model
+        identity :name
+        attribute :topic
+        attribute :push_config, :aliases => "pushConfig"
+        attribute :ack_deadline_seconds, :aliases => "ackDeadlineSeconds"
+
+        # Pulls from this subscription, returning any available messages. By
+        # default, this method returns immediately with up to 10 messages. The
+        # option 'return_immediately' allows the method to block until a
+        # message is received, or the remote service closes the connection.
+        #
+        # Note that this method automatically performs a base64 decode on the
+        # 'data' field.
+        #
+        # @param options [Hash] options to modify the pull request
+        # @option options [Boolean] :return_immediately If true, method returns
+        #   after checking for messages. Otherwise the method blocks until one
+        #   or more messages are available, or the connection is closed.
+        # @option options [Number] :max_messages maximum number of messages to
+        #   receive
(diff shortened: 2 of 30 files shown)


selmanj:
cc @erjohnso @Temikus @icco

lib/fog/google/models/pubsub/received_message.rb
```diff
@@ -0,0 +1,39 @@
...
+        # @see https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/acknowledge
+        def acknowledge
+          requires :ack_id
+
+          service.acknowledge_subscription(@subscription_name, [ack_id])
+          nil
+        end
+
+        def reload
+          # Message is immutable - do nothing
```
plribeiro3000: Maybe you could add a Log warning instead?

lib/fog/google/models/pubsub/subscription.rb
```diff
@@ -0,0 +1,90 @@
+require "base64"
+require "fog/core/model"
+require "fog/google/models/pubsub/received_message"
```
plribeiro3000: You might not need those extra requires here. Could you confirm you need them please?

plribeiro3000:
@selmanj LGTM. I'm not that into `google` cloud api but mostly my concerns are about following the conventions so we will be sure of not adding regressions like fog/fog#3430.

You followed everything so we are good here. 👍 

But i will let the other guys take a look at the actual implementation because they will have more insights.

Thanks!

icco:
Please add to the Readme as well.

lib/fog/google/models/pubsub/subscription.rb
selmanj: I took out all three and the tests/examples passed.  I don't think they are required.

I'm curious though why we should remove them - my understanding was that require wouldn't load a file more than once (I'm still relatively new to ruby so I am likely missing something obvious)

lib/fog/google/models/pubsub/received_message.rb
selmanj: Great idea!  Will do

lib/fog/google/models/pubsub/subscription.rb
plribeiro3000: Yeah. The thing about it is that we don't need it anymore thus it is kind of bogus and a legacy of when this was still inside the main `fog` gem.

So mostly of the approaches and stuff we are doing is move to a solution that was needed inside the main `fog` gem to a solution more accurate to a standalone gem. e.g `autoload`.

selmanj:
Updated README (hope it's worded alright)

selmanj:
@plribeiro3000 Thanks for the review!

selmanj:
I had to make a dependency change to enforce google-api-client on 0.8.6; unfortunately there is a bug in how google-api-client works on discovery documents that have a ':' in the path (which pubsub does)

See: https://github.com/google/google-api-ruby-client/commit/94a979b5126e6bb4e4a1779e512131ffe32525c1

icco:
Oh jeezus. That's weird. Ok, well I guess when we upgrade to 0.9 we'll be saved from this. Could you please in the pubsub section mention that you have to have a version > 0.8.5 to use pubsub, because someone will fuck that up... I'm nearly positive.

selmanj:
@icco Would it be better to cut a new major version with a required 0.8.6 dependency, or to continue allowing a range of dependencies but documenting in the pubsub area that previous versions won't work?

Personally I am leaning toward a new major version being less confusing.

icco:
I'd do both, document why we're requiring the dependency, and do a version bump. But because of how fog dependencies work, people won't always notice. It's a production dependency that we can only put in dev because otherwise it makes all of Fog require it.

lib/fog/google/models/pubsub/subscriptions.rb
```diff
@@ -0,0 +1,33 @@
...
+          data = service.list_subscriptions.body["subscriptions"] || []
+          load(data)
+        end
+
+        # Retrieves a subscription by name
+        #
+        # @param subscription_name [String] name of subscription to retrieve
+        # @return [Fog::Google::Pubsub::Topic] topic found, or nil if not found
+        def get(subscription_name)
+          if subscription = service.get_subscription(subscription_name).body
```
Temikus: Conditional assignment. This is confusing and will always get called if `service.get_subscription(subscription_name).body` doesn't return an exception.

lib/fog/google/models/pubsub/topics.rb
```diff
@@ -0,0 +1,32 @@
...
+          data = service.list_topics.body["topics"] || []
+          load(data)
+        end
+
+        # Retrieves a topic by name
+        #
+        # @param topic_name [String] name of topic to retrieve
+        # @return [Fog::Google::Pubsub::Topic] topic found, or nil if not found
+        def get(topic_name)
+          if topic = service.get_topic(topic_name).body
```
Temikus: Ditto for conditional assignment.

Temikus:
Thanks for this @selmanj! 

I have 2 minor comments on conditional assignment + please address @icco's comment on prod dependencies - otherwise LGTM

lib/fog/google/models/pubsub/subscriptions.rb
selmanj: You're right, it's much cleaner without.

lib/fog/google/models/pubsub/topics.rb
selmanj: Done.

selmanj:
@Icco I updated the README as instructed and bumped the major version

icco:
Cool!

erjohnso:
Woot, nice work @selmanj!

Thanks for the reviews @icco, @Temikus, @plribeiro3000! 💯

PR #120, opened 2016-04-23
Auto-Rubocop

icco:
Just cleaning and scrubbing.

### DIFF
--- a/lib/fog/compute/google/models/operations.rb
+++ b/lib/fog/compute/google/models/operations.rb
@@ -5,24 +5,24 @@ module Fog
         model Fog::Compute::Google::Operation
 
         def all(filters = {})
-          if filters["zone"]
-            data = service.list_zone_operations(filters["zone"]).body
-          elsif filters["region"]
-            data = service.list_region_operations(filters["region"]).body
-          else
-            data = service.list_global_operations.body
-          end
+          data = if filters["zone"]
+                   service.list_zone_operations(filters["zone"]).body
+                 elsif filters["region"]
+                   service.list_region_operations(filters["region"]).body
+                 else
+                   service.list_global_operations.body
+                 end
           load(data["items"] || [])
         end
 
         def get(identity, zone = nil, region = nil)
-          if !zone.nil?
-            response = service.get_zone_operation(zone, identity)
-          elsif !region.nil?
-            response = service.get_region_operation(region, identity)
-          else
-            response = service.get_global_operation(identity)
-          end
+          response = if !zone.nil?
+                       service.get_zone_operation(zone, identity)
+                     elsif !region.nil?
+                       service.get_region_operation(region, identity)
+                     else
+                       service.get_global_operation(identity)
+                     end
           return nil if response.nil?
           new(response.body)
         rescue Fog::Errors::NotFound
--- a/lib/fog/compute/google/requests/get_server.rb
+++ b/lib/fog/compute/google/requests/get_server.rb
@@ -50,11 +50,11 @@ module Fog
 
       class Real
         def get_server(server_name, zone_name)
-          if zone_name.is_a? Excon::Response
-            zone = zone_name.body["name"]
-          else
-            zone = zone_name
-          end
+          zone = if zone_name.is_a? Excon::Response
+                   zone_name.body["name"]
+                 else
+                   zone_name
+                 end
 
           api_method = @compute.instances.get
           parameters = {
--- a/lib/fog/compute/google/requests/insert_disk.rb
+++ b/lib/fog/compute/google/requests/insert_disk.rb
@@ -29,14 +29,14 @@ module Fog
           data[:disks][disk_name] = object
 
           if image
-            object.merge!("sourceImage" => image.self_link,
-                          "sourceImageId" => image.id)
+            object["sourceImage"] = image.self_link
+            object["sourceImageId"] = image.id
           end
-          if disk_type = options.delete(:type)
-            object["type"] = type
-          else
-            object["type"] = "https://www.googleapis.com/compute/#{api_version}/projects/#{@project}/zones/#{zone_name}/diskTypes/pd-standard"
-          end
+          object["type"] = if disk_type = options.delete(:type)
+                             type
+                           else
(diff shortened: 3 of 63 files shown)


lib/fog/compute/google/requests/insert_disk.rb
```diff
@@ -29,14 +29,14 @@ def insert_disk(disk_name, zone_name, image_name = nil, options = {})
...
-                          "sourceImageId" => image.id)
-          end
-          if disk_type = options.delete(:type)
-            object["type"] = type
-          else
-            object["type"] = "https://www.googleapis.com/compute/#{api_version}/projects/#{@project}/zones/#{zone_name}/diskTypes/pd-standard"
+            object["sourceImage"] = image.self_link
+            object["sourceImageId"] = image.id
           end
+          object["type"] = if disk_type = options.delete(:type)
```
icco: This is weird code... I can't quite decipher what this should be doing.

Temikus: @icco Looking diagonally, this seems to be an artifact of this: https://github.com/fog/fog/commit/c23eb01c8d57b295ef2cad0ba6aa8c23f79aa7a8

Theorising here: it was probably conceived as an easier way to set disk_type (without resorting to `get_disk_type` first), but probably was dismissed when the author saw that disk_types are not equally distributed between the zones.

BTW I wouldn't accept an autofix like that, it makes the code even more unreadable IMO :/

Temikus:
I want to raise one small concern about autofixing. Last time we did it - it made `git blame` totally unreadable. I think maybe we should consider going the "gradual changes" route with Hound and slowly overwriting the codebase?

@icco, @plribeiro3000 What do you guys think?

icco:
I don't mind the blame issues, as none of them are accurate to begin with (since we moved from fog -> fog-google).

Temikus:
I wouldn't say it isn't useful - I can at least see what is new vs. forklifted code, etc and we already have almost a year of changes in here.

Anyways, that's beside the point - It's fine now, I just think we shouldn't be doing this all the time on the whole codebase. This is at least the 8th time already:
```
λ git log | grep 'rubocop -a'
    rubocop -a
    rubocop -a
    rubocop -a
    rubocop -a on a bunch of files to appease hound
    rubocop -a
    rubocop -a
    rubocop -a
```

plribeiro3000:
@Temikus IT seems that @houndci-bot and `rubocop` are off.
They are not following the same styles. Thats probably why we need to fix it from time to time.
Maybe we should stick to one or another?

Temikus:
@plribeiro3000 Huh, I thought it just runs through rubocop.yml
So, it just skips some rules in there or...?

icco:
You have to tell hound to use robocop, otherwise it uses it's own style rules

plribeiro3000:
@icco It seems we are doing this: https://github.com/fog/fog-google/blob/master/.hound.yml

Unless it changed somehow.

icco:
hmm, awkward. I do not know.

lib/fog/compute/google/models/operations.rb
```diff
@@ -5,24 +5,24 @@ class Operations < Fog::Collection
...
 
         def get(identity, zone = nil, region = nil)
-          if !zone.nil?
-            response = service.get_zone_operation(zone, identity)
-          elsif !region.nil?
-            response = service.get_region_operation(region, identity)
-          else
-            response = service.get_global_operation(identity)
-          end
+          response = if !zone.nil?
```
Temikus: Is this really idiomatic? IMO this would be ok with a `when` statement, but it seems wrong to me. 
What do you think?

lib/fog/compute/google/models/operations.rb
```diff
@@ -5,24 +5,24 @@ class Operations < Fog::Collection
...
 
         def all(filters = {})
-          if filters["zone"]
-            data = service.list_zone_operations(filters["zone"]).body
-          elsif filters["region"]
-            data = service.list_region_operations(filters["region"]).body
-          else
-            data = service.list_global_operations.body
-          end
+          data = if filters["zone"]
```
Temikus: Ditto ^

lib/fog/compute/google/requests/get_server.rb
```diff
@@ -50,11 +50,11 @@ def get_server(server_name, zone_name)
 
       class Real
         def get_server(server_name, zone_name)
-          if zone_name.is_a? Excon::Response
-            zone = zone_name.body["name"]
-          else
-            zone = zone_name
-          end
+          zone = if zone_name.is_a? Excon::Response
```
Temikus: Ditto^

tests/helper.rb
```diff
@@ -100,12 +86,10 @@ def create_test_target_pool(connection, region)
...
   http_health_check = create_test_http_health_check(connection)
   instance = create_test_server(connection, "us-central1-a")
-  target_pool = connection.target_pools.create({
-                                                 :name => "fog-test-target-pool-#{random_string}",
+  target_pool = connection.target_pools.create(                                                 :name => "fog-test-target-pool-#{random_string}",
                                                  :region => region,
                                                  :health_checks => [http_health_check.self_link],
-                                                 :instances => [instance.self_link]\
-                                               })
+                                                 :instances => [instance.self_link]\)
```
Temikus: Please fix the incorrect backslash, it's breaking the build.

Temikus:
I think there's no discrepancy, we just need to be better at actually making sure people fix all Hound complaints or, if the linter is unapplicable, put a rubocop override (for example `# rubocop:disable Metrics/ClassLength`)
Like the JSON PR has at 20+ unfixed complaints: https://github.com/fog/fog-google/pull/96
This PR has 3 too ;)

@icco Anyways, just take a look at my comments - otherwise LGTM.

Temikus:
@icco Just to clarify - why did you close this PR? I hope I wasn't too naggy... 😞

icco:
I just figured I'd do it as we change other things, there were merge conflicts and I didn't want to deal with it.

Temikus:
Understood, thanks for making it clear :)

PR #122, opened 2016-04-24
Broken signing for GCS XML

### LINKED ISSUE #121: GCS Signatures not working
I'm still digging into this, but I think a change we've made recently has broken request signing to GCS.

icco:
Fixes #121. 

Any thoughts on how to test?

### DIFF
--- a/lib/fog/google/version.rb
+++ b/lib/fog/google/version.rb
@@ -1,5 +1,5 @@
 module Fog
   module Google
-    VERSION = "0.3.0".freeze
+    VERSION = "0.3.1".freeze
   end
 end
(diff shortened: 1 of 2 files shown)


plribeiro3000:
Well. Every time i don't know how to test something i just try to write it in a way i can.
But since this is really legacy at this point and really huge i would say that if you tried it yourself and is working is good to go.

But we need to refactor this at some point.

icco:
Yeah, I'm using this in https://github.com/icco/wallpapers/blob/master/Rakefile#L39, but I'll think about how to refactor and test.

Issue #124, opened 2016-05-20
getting NoMethodError: undefined method `keys' for nil:NilClass


jiggneshhgohel:
**Ruby: 2.3.0**
**Rails: 4.2.5**
**Paperclip: 4.3.6**
**fog-google: 0.3.1**

In my Rails app I am using Paperclip for file uploads and using Paperclip's Fog storage for 
uploading the files to Google Cloud.

I am facing an error `NoMethodError: undefined method `keys' for nil:NilClass` **while trying to update an already uploaded image file** on my Google Cloud Storage Bucket

**Photo** model:

```
class Photo < ActiveRecord::Base
  belongs_to :imageable, polymorphic: true

  has_attached_file :photo

  validates_attachment_presence :photo
  validates_attachment_content_type :photo, content_type: /\Aimage/
  validates_attachment_file_name :photo, matches: [/png\Z/, /jpe?g\Z/]
  validates_attachment_size :photo, less_than_or_equal_to: 2.megabytes

  validates_presence_of :imageable

  validates_uniqueness_of :imageable_id, scope: :imageable_type, message: 'photo already exists', case_sensitive: false

  def url(photo_style: nil)
    case photo_style
      when :medium
        photo.url(:medium)
      when :thumbnail
        photo.url(:thumbnail)
      else
        photo.url
    end
  rescue Exception => e
    Rails.logger.error "Exception occurred while trying to fetch url for Photo(##{self.id}). Exception message: #{e.message}"
    Rails.logger.error "Returning Paperclip's default_url for Photo(##{self.id})."
    photo.options[:default_url]
  end
end
```

Below is the output of my rails console which shows the stack trace pointing to line `fog-google-0.3.1/lib/fog/storage/google_xml/real.rb:66` in this gem:

```
Loading development environment (Rails 4.2.5)
2.3.0 :001 > p = Photo.find(9)
  Photo Load (0.8ms)  SELECT  "photos".* FROM "photos" WHERE "photos"."id" = $1 LIMIT 1  [["id", 9]]
 => #<Photo id: 9, imageable_type: "Profile", imageable_id: 19, created_at: "2016-05-20 14:19:46", updated_at: "2016-05-20 14:19:46", photo_file_name: "2.png", photo_content_type: "image/png", photo_file_size: 1764904, photo_updated_at: "2016-05-20 14:19:45">
2.3.0 :002 > p.photo
[...]

PR #126, opened 2016-05-23
Fix for XML storage issues

Temikus:
- Fixed more path escape issues on Storage XML, didn't update everything in https://github.com/fog/fog-google/pull/83
- Fixed regression in query enumeration in `signature` method.
- Added additional line to examples using the method that failed.
Fixing #124 

/CC @plribeiro3000 or @icco PTAL


icco:
LGTM!

plribeiro3000:
:+1:

Issue #132, opened 2016-06-01
Instance groups model semantics are incorrect


Temikus:
@areznikov-kabam seems to be right in #115 , `instance_groups` methods include:
```
def add_instance(params)
def remove_instance(params)
def get_instances(params)
```
, which should be in `instance_group` model according to current structure.

Since people are relying on those already, I propose either:
- A change in the next major version.
or/and
- Add the correct methods to `instance_group` model too and add deprecation warnings to  `instance_groups`.

@icco @plribeiro3000 - WDYT?

plribeiro3000:
Does the ones at `instance_groups` work at all?

If the don't just remove them and add in the right place otherwise i guess for now we should add the methods on the right place and deprecate the wrong ones.

Temikus:
@plribeiro3000 That's the problem, the ones in `instance_groups` **do** work. So we'll need to deprecate them somehow eventually. I deprecated only the `get_instances` from the collection since it wasn't working anyway.

plribeiro3000:
Hmmm. Lets deprecate them. And add the correct ones.
Then we do a minor release and on a Major one we remove the deprecated apis.

Temikus:
@plribeiro3000 I already did ;) See the #133

plribeiro3000:
👍

Temikus:
Fixed by #133

Issue #115, opened 2016-04-05
instance_groups.all() fails with TypeError: no implicit conversion of String into Integer


areznikov-kabam:
fog-google 0.2.0
google-api-client 0.8.6
RUBY_VERSION "2.0.0"

instance_groups.all() fails with the following error

```
>> gce.instance_groups.all()
Exception `OpenSSL::SSL::SSLError' at /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/openssl/buffering.rb:175 - read would block
Exception `OpenSSL::SSL::SSLError' at /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/openssl/buffering.rb:175 - read would block
Exception `TypeError' at /Library/Ruby/Gems/2.0.0/gems/fog-google-0.2.0/lib/fog/compute/google/models/instance_groups.rb:17 - no implicit conversion of String into Integer
Exception `TypeError' at /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/irb/workspace.rb:86 - no implicit conversion of String into Integer
TypeError: no implicit conversion of String into Integer
	from /Library/Ruby/Gems/2.0.0/gems/fog-google-0.2.0/lib/fog/compute/google/models/instance_groups.rb:17:in `[]'
	from /Library/Ruby/Gems/2.0.0/gems/fog-google-0.2.0/lib/fog/compute/google/models/instance_groups.rb:17:in `all'
	from (irb):118
	from /usr/bin/irb:12:in `<main>'
```

areznikov-kabam:
I've fixed it by changing one line
17c17
<           load(data)
---
>           load(data["items"])

Anyway lib/fog/compute/google/models/instance_groups.rb seems to me a little bit wrong
def add_instance(params)
def remove_instance(params)
def get_instances(params)
should be moved to lib/fog/compute/google/models/instance_group.rb following my understanding and improved to something similar to lib/fog/compute/google/models/target_pool.rb

Temikus:
Fixed by #133 
Thanks for submitting this, @areznikov-kabam! 🎉

PR #133, opened 2016-06-01
Fixing some instance_group issues

Temikus:
- Moving methods to the correct model
- Cleaned up requests
- Renaming get_instances to be coherent with GCE API
- Refactoring InstanceGroup methods to support multiple instances	

Fixing #115, #132

@plribeiro3000 PTAL, please, if possible make sure I'm symantically correct, since this is public-facing API (see #132)


lib/fog/compute/google/models/instance_groups.rb
```diff
@@ -32,17 +32,15 @@ def get(identity, zone = nil)
           nil
         end
 
+        # TODO: This shouldn't be in this model, add deprecation warning
         def add_instance(params)
```
plribeiro3000: I think you can use `Fog` api to add this deprecation. Is `Fog.deprecate` if i'm not mistaken.

icco: Yup, you can, and it'll complain in all the right ways.

icco: Another thing you could do is just call the new undeprecated function.

Temikus:
Added Fog::Logger.deprecation messages to the method. It's still using the request directly since I'm a bit wary of instantiating new model objects inside the collection. Do tell me if that is unfounded though.

@plribeiro3000 PTAL

lib/fog/compute/google/models/instance_group.rb
```diff
@@ -27,6 +27,24 @@ def destroy(_async = true)
           data = service.delete_instance_group(name, zone_name)
         end
 
+        def add_instance(instance_name)
```
plribeiro3000: Do you need the signature to be `add_instance`?
The object name is `Instance` so maybe just `add` is enough?

Another point is that perhaps this should be the create method of the `Instance` collection?

Temikus: Sorry but I'm confused :/ The class is `InstanceGroup`, no?
In this case "add_instance" is called because instance is being added to an instance group, this is not a collection, but rather GCE's own groups: https://cloud.google.com/compute/docs/reference/latest/instanceGroups/addInstances

Or do you mean extending `Instance` somehow?

plribeiro3000: Ouch. I get it now. You have a `InstanceGroup` object and you are adding `Instance`'s to that object.
So GCE does abstract the collections into its own resource (`InstanceGroup`). I did not know that.

👍 then.

plribeiro3000:
Just by curiosity. Do we have an `InstanceGroup` collection object form `fog`?

Temikus:
@plribeiro3000 While we're at it, I noticed that the actual call in GCE is adding an array of instances to be included in the group, rather than a single instance like we have now:
https://cloud.google.com/compute/docs/reference/latest/instanceGroups/addInstances

Should I think about renaming it to `add_instances` and writing a new request to go with it? (While we're at it anyways).

However, for that I would probably need to rewrite the whole request :/

Temikus:
(on InstanceGroup collection)
Seems `cloudstack` has one at least:
```
λ ack 'InstanceGroup' .
lib/fog/cloudstack/requests/compute/create_instance_group.rb
8:        # {CloudStack API Reference}[http://cloudstack.apache.org/docs/api/apidocs-4.4/root_admin/createInstanceGroup.html]
13:            options.merge!('command' => 'createInstanceGroup')
15:            options.merge!('command' => 'createInstanceGroup',

lib/fog/cloudstack/requests/compute/delete_instance_group.rb
8:        # {CloudStack API Reference}[http://cloudstack.apache.org/docs/api/apidocs-4.4/root_admin/deleteInstanceGroup.html]
13:            options.merge!('command' => 'deleteInstanceGroup')
15:            options.merge!('command' => 'deleteInstanceGroup',

lib/fog/cloudstack/requests/compute/list_instance_groups.rb
8:        # {CloudStack API Reference}[http://cloudstack.apache.org/docs/api/apidocs-4.4/root_admin/listInstanceGroups.html]
13:            options.merge!('command' => 'listInstanceGroups')
15:            options.merge!('command' => 'listInstanceGroups')

lib/fog/cloudstack/requests/compute/update_instance_group.rb
8:        # {CloudStack API Reference}[http://cloudstack.apache.org/docs/api/apidocs-4.4/root_admin/updateInstanceGroup.html]
13:            options.merge!('command' => 'updateInstanceGroup')
15:            options.merge!('command' => 'updateInstanceGroup',
```

Temikus:
Discussed with Paulo offline - please, don't merge yet, I'll refactor the requests so we don't have to deal with this later.

lib/fog/compute/google/models/instance_group.rb
```diff
@@ -27,6 +27,25 @@ def destroy(_async = true)
...
+          service.add_instance_group_instance(identity, zone_name, instance_name)
+        end
+
+        def remove_instance(instance_name)
+          requires :identity, :zone
+
+          service.remove_instance_group_instance(identity, zone_name, instance_name)
+        end
+
+        def list_instances
```
Temikus: This will return:
```
=> [{"instance"=>"https://www.googleapis.com/compute/v1/projects/dev-utility-90115/zones/europe-west1-d/instances/chef-server", "status"=>"RUNNING"},
 {"instance"=>"https://www.googleapis.com/compute/v1/projects/dev-utility-90115/zones/europe-west1-d/instances/elk-server", "status"=>"RUNNING"}]
```
@icco @plribeiro3000  - PTAL - is such a return acceptable or should I return an array of self-links?

Temikus: Additionally, if there's no data it passes back `nil` - is that ok or should it be empty array?

Temikus:
@plribeiro3000 PTAL, I think I'm done with this.

lib/fog/compute/google/models/instance_group.rb
```diff
@@ -27,9 +27,39 @@ def destroy(_async = true)
...
+
+        def remove_instances(instances)
+          requires :identity, :zone
+
+          service.remove_instance_group_instances(
+            identity, zone_name, format_instance_list(instances)
+          )
+        end
+
+        def list_instances
```
Temikus: This will return:
```
=> [{"instance"=>"https://www.googleapis.com/compute/v1/projects/dev-utility-90115/zones/europe-west1-d/instances/chef-server", "status"=>"RUNNING"},
 {"instance"=>"https://www.googleapis.com/compute/v1/projects/dev-utility-90115/zones/europe-west1-d/instances/elk-server", "status"=>"RUNNING"}]
```
@plribeiro3000 - PTAL - is such a return acceptable or should I return an array of self-links?
Additionally, if there's no data it passes back nil - is that ok or should it be empty array?

lib/fog/compute/google/models/instance_group.rb
```diff
@@ -27,9 +27,39 @@ def destroy(_async = true)
...
+          data["items"]
+        end
+
         def zone_name
           zone.nil? ? nil : zone.split("/")[-1]
         end
+
+        private
+
+        def format_instance_list(instance_list)
```
Temikus: @plribeiro3000 - this private method ensures that the model method can accept instance models as input (I don't want to overload the request with that logic). Is this acceptable?

lib/fog/compute/google/models/instance_group.rb
icco: `InstanceGroup` is a type of grouping of instances. You can use it do load balance and auto-scaling: https://cloud.google.com/compute/docs/instance-groups/

lib/fog/compute/google.rb
```diff
@@ -18,153 +18,153 @@ class Google < Fog::Service
...
-      request :set_server_disk_auto_delete
-      request :set_server_scheduling
-      request :add_server_access_config
-      request :delete_server_access_config
-      request :update_url_map
-      request :validate_url_map
       request :start_server
       request :stop_server
+      request :update_url_map
+      request :validate_url_map
```
icco: 👍 yay sorting!

lib/fog/compute/google/models/instance_group.rb
icco: Since this is a model, I would assume this would return instances of the `Instance` object.

Temikus: @icco Understood, I'll refactor this.

lib/fog/compute/google.rb
Temikus: Well, it was in a bit of a mess, so I decided to clean the list up a bit, why not :-P

Temikus:
TODO:
- [ ] Change `list_instances` to produce an array of `instance` models.

Temikus:
@icco I refactored the `list_instances` to return an array of `Server` models, but I'm not sure whether the way I coded it was the most efficient available. PTAL.

Output:
```
connection = Fog::Compute.new({ :provider => "Google" })
ig = connection.instance_groups.get("devops-infra")

ig.list_instances
=> [  <Fog::Compute::Google::Server
    name="chef-server",
    kind="compute#instance",
    ...REDACTED... 
  >,
   <Fog::Compute::Google::Server
    name="elk-server",
    kind="compute#instance",
    ...REDACTED...   
  >]
[5] pry(main)>
```

lib/fog/compute/google/models/instance_group.rb
plribeiro3000: @Temikus @icco Agreed. You might be able to make it easier using the `Fog` DSL for associations.

lib/fog/compute/google/models/instance_group.rb
plribeiro3000: IMHO this is right. In the past we had some requests that were overloaded with such kind of logic and it was a PITA because test it is hard since you gotta mock the actual request somehow.

plribeiro3000: What i think we can do here is create an abstraction model that all other model inherits.
Then this method can be reused in other models if needed. I did this on `fog-xenserver` for the shared logic.

lib/fog/compute/google/models/instance_group.rb
```diff
@@ -18,18 +18,54 @@ class InstanceGroup < Fog::Model
...
         end
 
         def zone_name
           zone.nil? ? nil : zone.split("/")[-1]
         end
+
+        private
+
+        def format_instance_list(instance_list)
+          instance_list = Array(instance_list)
```
plribeiro3000: 👍 

This is a feature that should be used more often.

lib/fog/compute/google/models/instance_group.rb
Temikus: @plribeiro3000 Thx! I'll think about abstracting it away when working on #134. So far we don't have many models accepting collections models as input as far as I've seen.

plribeiro3000: 👍

Temikus:
@icco PTAL, addressed your comments. 
Otherwise - I think it's ready to be merged.

icco:
:+1:

Temikus:
Merging 🎉  Thanks @icco and @plribeiro3000 for the review!

PR #137, opened 2016-07-11
Remove the default value of "MIGRATE" for "onHostMaintenance"

### LINKED ISSUE #136: Provisioning a server with the :preemptible option requires specifying :on_host_maintenance as well
By default, the option for `:on_host_maintenance` is set to `MIGRATE`. However, if someone attempts to create a preemptible vm via `:preemptible`, then `:on_host_maintenance` must be set to `TERMINATE`. This is slightly surprising given that the google API documents the `onHostMaintanence` setting as:

> Defines the maintenance behavior for this instance. For standard instances, the default behavior is MIGRATE. For preemptible instances, the default and only possible behavior is TERMINATE.

https://cloud.google.com/compute/docs/reference/latest/instances#resource-representations

Given that this parameter is optional and by default set to the correct value, likely fog-google shouldn't specify it unless the request does.

selmanj:
This value is incompatible with the default used by the GCE api;
specifically, the value 'MIGRATE' is incompatible with pre-emptible
instances. Fog-google should instead respect the default value GCE
chooses rather than imposing its own.

Fixes #136

### DIFF
--- a/lib/fog/compute/google/requests/insert_server.rb
+++ b/lib/fog/compute/google/requests/insert_server.rb
@@ -161,6 +161,10 @@ module Fog
           if options.key? "preemptible"
             scheduling["preemptible"] = options.delete "preemptible"
             scheduling["preemptible"] = scheduling["preemptible"].class == TrueClass
+            # If we don't have an explicit value for "on_host_maintenance, then
+            # force the value to be 'TERMINATE' as it's the default and
+            # only-accepted value for preemptible vms
+            scheduling["onHostMaintenance"] = "TERMINATE" unless options.key? "on_host_maintenance"
           end
           if options.key? "on_host_maintenance"
             ohm = options.delete "on_host_maintenance"


selmanj:
Arg, looks like the build is now failing due to a transitive dependency on activesupport 5.0.0.  Looks like this is getting pulled in from the older version of google-api-client (the dependency on activesupport was removed in 0.9).

Does this mean users on version of ruby older than 2.2.2 are broken now?

Temikus:
@selmanj - created an issue for the build failing, PTAL since it relates to your previous PR: #138

About this one - I do agree with not imposing things, however, I would rather keep it as it until the next major version since this may break existing behaviour. 
Does creating an instance with `preemptible` option actually succeeds with your changes without explicitly stating the scheduling to `TERMINATE` in this case?

selmanj:
This change shouldn't be breaking as it's relaxing a restriction.  If you don't specify a value for `onHostMaintenance` the GCE api will choose one for you. From https://cloud.google.com/compute/docs/reference/latest/instances:

```
Defines the maintenance behavior for this instance. For standard instances, the default behavior is MIGRATE. For preemptible instances, the default and only possible behavior is TERMINATE. For more information, see Setting Instance Scheduling Options.
```

Without this change, a user creating a preemptible instance has to explicitly set `onHostMaintenance` to `TERMINATE`; otherwise they receive an invalid request error.  After this change, a user can explicitly set `onHostMaintenance` or they can leave it unset.

selmanj:
Closing and reopening PR to force a rebuild on travis.

icco:
So I don't like this, I'd much rather you overrode the default setting when you set preemptible. GCE wants people to set MIGRATE by default, otherwise there will probably be far more user outages than a user expects.

selmanj:
@icco By forcing users to override the default setting, the fog api will then be different than the public GCP api (which, if `onHostMaintenance` is not specified, uses `TERMINATE` if the host is preemptible, otherwise uses `MIGRATE`).  I think this would be confusing to users and it's not clear why fog would insist on this (especially given that if the vm is preemptible, you MUST specify `TERMINATE` as migration isn't supported).

icco:
@selmanj What I'm saying is your code doesn't do that.

There is a whole block of code dealing with preemptible right below this.

icco:
This is how I'd deal with this:

```ruby
# Defaults
scheduling = {
  "automaticRestart" => false,
  "onHostMaintenance" => "MIGRATE",
  "preemptible" => false,
}

if options.key? "auto_restart" && options["auto_restart"]
  scheduling["automaticRestart"] = !!options["auto_restart"]
end

if options.key? "preemptible" && options["preemptible"]
  scheduling["preemptible"] = !!options["preemptible"]
  scheduling["onHostMaintenance"] = "TERMINATE"
end

if options.key? "on_host_maintenance"
  scheduling["onHostMaintenance"] = options["on_host_maintenance"].to_s.upcase
end

body_object["scheduling"] = scheduling
```

selmanj:
@icco Right that would work, but the API will do the same thing anyway if we don't set a value, so why the additional code?

icco:
Because the API changes over time and I prefer to be explicit. For example: Migrate wasn't always the default (there was a time where there was no default and you had to be explicit).

selmanj:
@icco I'm not opposed to making that change. Will revise and resubmit

icco:
Thanks!

PR #139, opened 2016-07-13
Lock activesupport dependency to <5 for backwards compatibility

### LINKED ISSUE #138: Activesupport/google-api-clien dependency issues
It seems that currently all builds before ruby `2.2.2` are failing due to activesupport `5.0.0`:

@selmanj - Looking at this diagonally this seems to be a `google-api-client` problem, changed when you added pubsub in #118 

Previous versions (before 0.8.5) used to have a proper pessimistic constraint on activesupport: https://rubygems.org/gems/google-api-client/versions/0.8.4

Can you tell me which version does your pubsub implementation rely on? Is it ok to roll back to 0.8.4?

selmanj:
Closes #138

### DIFF
--- a/fog-google.gemspec
+++ b/fog-google.gemspec
@@ -28,6 +28,11 @@ Gem::Specification.new do |spec|
 
   # TODO: Upgrade to 0.9, which is not compatible.
   spec.add_development_dependency "google-api-client", "~> 0.8.6"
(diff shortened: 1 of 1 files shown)


Temikus:
@selmanj Thanks!

FYI: I created an issue with ustream. It's a long shot, but still: https://github.com/google/google-api-ruby-client/issues/438

PR #143, opened 2016-07-21
Allow additional options to be applied to body

### LINKED ISSUE #142: Unable to specify a lifecycle configuration when creating a bucket
https://cloud.google.com/storage/docs/json_api/v1/buckets/insert
https://cloud.google.com/storage/docs/lifecycle

There's no way to specify additional options in the body when creating a bucket; only parameter options are supported: https://github.com/fog/fog-google/blob/master/lib/fog/storage/google_json/requests/put_bucket.rb#L17

I'm guessing the reason for the lack of options is because the Amazon API has a specific lifecycle resource that you use rather than the bucket. Probably the simplest thing to do is allow additional options to be attached to the body.  

/cc @Temikus

selmanj:
Fixes #142.

/cc @Temikus 

Note that I'm not entirely sure that this is the best way to solve this (and would welcome any discussion). Adding an additional optional parameter allows us to not break existing implementations, but it feels pretty free-form. I'm not sure it's beneficial to perform some sort of whitelist or something though as it will only restrict flexibility of the client in the future.

Thoughts?

### DIFF
--- a/lib/fog/storage/google_json/requests/put_bucket.rb
+++ b/lib/fog/storage/google_json/requests/put_bucket.rb
@@ -10,11 +10,12 @@ module Fog
         # * options<~Hash> - config arguments for bucket.  Defaults to {}.
         #   * 'LocationConstraint'<~Symbol> - sets the location for the bucket
         #   * 'x-amz-acl'<~String> - Permissions, must be in ['private', 'public-read', 'public-read-write', 'authenticated-read']
+        # * body_options<~Hash> - body arguments for bucket creation
         #
         # ==== Returns
         # * response<~Excon::Response>:
         #   * status<~Integer> - 200
-        def put_bucket(bucket_name, options = {})
+        def put_bucket(bucket_name, options = {}, body_options = {})
           location = options["LocationConstraint"] if options["LocationConstraint"]
 
           api_method = @storage_json.buckets.insert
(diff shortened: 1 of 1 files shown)


Temikus:
@selmanj Do I read correctly that lifecycle management can be set through API but cannot be enabled without `gsutil`? Doesn't seem good to tell users to perform manual steps with third-party utilities along with fog.

Is this the only feature that needs additional body options?

selmanj:
@Temikus Where did you read that?  I believe that the lifecycle config is applied when it's specified on creation of the bucket (or on later updating of the bucket via PATCH)

Temikus:
@selmanj https://cloud.google.com/storage/docs/lifecycle#setup
Though, reading through it again, it seems that they may have just forgotten about the API route... My miss, nevermind that then.

A change like this would indeed allow for the easiest way to change bucket lifecycle, logging and website options for people who want to do that. We can also add more user-friendly options later if we want to.

My main concern is not to introduce something people may rely on and we can't deprecate/rewrite easily, so I want to have another opinion on this.

@icco - WDYT? Does this look like something that may bite us later? It doesn't look so to me but I want to have your opinion on this.

icco:
So based on reading https://cloud.google.com/storage/docs/json_api/v1/buckets/insert, this makes a lot of sense. What is confusing, is we're going to have to do weird things to make the xml api similar (https://cloud.google.com/storage/docs/xml-api/put-bucket-lifecycle).

icco:
So @Temikus, I'd say this is fine, but, when we add lifecycle management to the models, it'll be a bunch of weird code.

Temikus:
LGTM. Thanks @selmanj for the PR and thanks to @icco for review!

PR #152, opened 2016-08-10
Make servers.get to search for identity in all parts of it's name

akarasik:
In a given environment of 2 instances named:
- 1test
- test1

If I execute servers.get("test"), I will not receive the "test1" object, but I will receive the "1test".

### DIFF
--- a/lib/fog/compute/google/models/servers.rb
+++ b/lib/fog/compute/google/models/servers.rb
@@ -21,7 +21,7 @@ module Fog
           if zone
             response = service.get_server(identity, zone).body
           else
-            servers = service.list_aggregated_servers(:filter => "name eq .*#{identity}").body["items"]
+            servers = service.list_aggregated_servers(:filter => "name eq .*#{identity}.*").body["items"]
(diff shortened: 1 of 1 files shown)


Temikus:
Hi @akarasik! Thanks for the contribution!

The danger of this change  here is that it may return not the instance you expect it to.

The `get` method exists to get only one singular server on the project. 
I don't think it should even have the `.*` in the beginning since it looks like a leftover from the code when instance names used to be specified as a full URL, like `https://www.googleapis.com/compute/v1/projects/all-hail-hypnotoad/zones/europe-west1-d/instances/elk-server`

With that said, however, I think it would be cool to just implement a new method called `search` or  `query` which would return a collection of servers matching the regex.

@icco or @plribeiro3000 - if you don't mind - what are your thoughts on this? You've been on the project longer than me.

akarasik:
> With that said, however, I think it would be cool to just implement a new method called search or query which would return a collection of servers matching the regex.

That's also what I thought, but then I saw that the get will "partially" search hosts.
A separate search\query method should be great!

Temikus:
Closing in favor of #181

Issue #153, opened 2016-08-11
Unable to specify proxy option when using any google client


selmanj:
The base google-api-client supports specifying a proxy option:

https://github.com/google/google-api-ruby-client/blob/v0.8.6/lib/google/api_client.rb#L110

fog-google just needs to be extended to expose it. Should be pretty easy to add.

Temikus:
Wait, it already does support proxy by grabbing the `http_proxy` variable from the environment. 
This is IMO a correct way to set those settings since prod proxies usually require some sort of credentials.

selmanj:
Is that the correct way?  How are the credentials set?

I believe the other fog providers handle the proxy option via the options array; an environment variable works but tends to be a lot less explicit.

Temikus:
@selmanj Already discussed in the #154 - they want to add a provider-specific proxy option, so this will not do indeed.

P.S. On how credentials are set usually:
```
export http_proxy='http://user:password@prox-server:3128'
export https_proxy='http://user:password@prox-server:3128'
export ftp_proxy='http://user:password@prox-server:3128'
```

selmanj:
@Temikus I don't think the proxy option is provider-specific; the AWS one for example accepts proxy. Adding the ability to specify the proxy in the options allows different proxy settings for different services, and tends to be more explicit than relying on environment variables.

I can't tell where you guys landed on #154 regarding this option, but I strongly feel that relying on just the environment variable is both hard-to-discover for users, and brittle.

Temikus:
@selmanj As I said in #154, I think yours is a good idea and I like that type of implementation since we won't need to provide a proxy option (next google-api-client may very well not support it, for example)

kgaikwad:
@Temikus, I would like to know what is pending for this change.

kgaikwad:
> Any suggestions on this?
Please ignore. Got the reference - https://github.com/fog/fog-google/pull/418#issuecomment-422848146

PR #154, opened 2016-08-12
[WIP] Adding Proxy support

juliancheal:
As mentioned in issue #153 the Google Api client supports proxies.
This is an inital work in progress at getting proxy support in fog-google.

This is my first look at the fog-google gem, so I'm not that familiar.

I know this PR isn't complete, but thought I'd make an initial attempt and ask for feedback on what exactly else needs doing to get proxy support into fog-google.

Thanks :)

### DIFF
--- a/lib/fog/google/shared.rb
+++ b/lib/fog/google/shared.rb
@@ -29,6 +29,7 @@ module Fog
       # @option options [String] :app_name The app name to set in the user agent
       # @option options [String] :app_version The app version to set in the user agent
       # @option options [Google::APIClient] :google_client Existing Google API Client
+      # @option options [String] :proxy The proxy to send requests
       # @return [Google::APIClient] Google API Client
       # @raises [ArgumentError] If there is any missing argument
       def initialize_google_client(options)
@@ -62,7 +63,8 @@ module Fog
           signing_key,
           options[:google_api_scope_url],
           options[:app_name],
-          options[:app_version]
+          options[:app_version],
+          options[:proxy]
         )
       end
 
@@ -109,7 +111,7 @@ module Fog
       # @param [String] app_name The app name to set in the user agent
       # @param [String] app_version The app version to set in the user agent
       # @return [Google::APIClient] Google API Client
-      def new_pk12_google_client(google_client_email, signing_key, google_api_scope_url, app_name = nil, app_version = nil)
+      def new_pk12_google_client(google_client_email, signing_key, google_api_scope_url, app_name = nil, app_version = nil, proxy = nil)
         application_name = app_name.nil? ? "fog" : "#{app_name}/#{app_version || '0.0.0'} fog"
         api_client_options = {
           :application_name => application_name,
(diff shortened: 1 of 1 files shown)


Temikus:
Hi @juliancheal ! Thanks for the PR! :)

I see that the option is already exposed in variable by setting `http_proxy` variable. Which is, IMO, correct since prod proxies usually include some sort of credentials.
Or am I missing some specific use-case for an argument?

juliancheal:
Hey @Temikus.

I'm trying to set the proxy settings from ManageIQ. This is the PR I'm currently working on https://github.com/ManageIQ/manageiq/pull/10368

Temikus:
@juliancheal Ah, makes sense now. Thanks for the reference!

Just a couple of notes then:
- Did you take a look what happens if the `http_proxy` is set? Would the parameter be overriden?
- Did you check whether it works the same with API client 0.9? We're going to migrate soon.

juliancheal:
> Did you take a look what happens if the http_proxy is set? Would the parameter be overridden?

Do you mean setting `http_proxy` as an environment variable. The problem with that is customers want to be able to set separate proxies for different providers, i.e. amazon, google, azure. So setting a global env wouldn't work in this case.

> Did you check whether it works the same with API client 0.9? We're going to migrate soon.

I've not, I'll give that a go, and make sure it works there too. Thanks for the heads up. :)

Temikus:
@juliancheal 
> Do you mean setting http_proxy as an environment variable. The problem with that is customers want to be able to set separate proxies for different providers, i.e. amazon, google, azure. So setting a global env wouldn't work in this case.

If you look closely `google-api-client` already implements the logic for taking the environment variable in and setting it as a proxy parameter.
What I meant is - make sure that the lib itself doesn't explode if you set both the `proxy:` in fog and have `http_proxy` set for any reason.

> I've not, I'll give that a go, and make sure it works there too. Thanks for the heads up. :)
 Cool, thanks 👍

juliancheal:
> If you look closely google-api-client already implements the logic for taking the environment variable in and setting it as a proxy parameter.
> What I meant is - make sure that the lib itself doesn't explode if you set both the proxy: in fog and have http_proxy set for any reason.

Ah, now I see what you mean. Thanks @Temikus :)

selmanj:
Instead of adding a `proxy` parameter, would it be better to add a `google_client_options` hash that will be directly passed to the underlying google client?  I'm not sure if fog-google needs to be aware of any of these options, and in the future it would allow consumers to make changes to the underlying client without having to 'poke a hole' like this PR does.

@Temikus what do you think?

Temikus:
@selmanj - That is a good idea. 

However, our stats depend on `application_` options in that hash, so one needs to make sure those are not overwritten.

juliancheal:
Sorry for my absence on this PR, have been away traveling.

I've updated this PR (if it were to be merged I will squish the commits) with the suggestion from @selmanj of creating a `google_client_options` hash.

Temikus:
Sorry it took a while. Thanks @juliancheal!

juliancheal:
Hi @Temikus no problem, thanks for merging 😄

PR #155, opened 2016-08-20
Removing activesupport workaround

Temikus:
Since 0.8.7 was released.
See https://github.com/google/google-api-ruby-client/issues/438


icco:
:+1:

PR #156, opened 2016-08-25
allow copy to accept an options parameter

zrob:
this creates parity with other providers such as AWS, allowing clients
to use S3 and Cloud Storage interchangeably

### DIFF
--- a/lib/fog/storage/google_json/models/file.rb
+++ b/lib/fog/storage/google_json/models/file.rb
@@ -52,9 +52,9 @@ module Fog
 
         attr_reader :directory
 
-        def copy(target_directory_key, target_file_key)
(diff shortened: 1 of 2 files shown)


dsboulder:
@erjohnso This PR will allow us to use GCS in CF cloud controller as our blobstore.

erjohnso:
Thanks for this @zrob - do you mind addressing the houndci comments? Otherwise, looks good to me and we'll get it merged.

zrob:
@erjohnso looks like the hounds are happy.  let me know if you need anything else.  thanks!

erjohnso:
Thanks @zrob and @icco!

dsboulder:
@icco when can we public a new version of the fog-google gem that includes this fix?

PR #164, opened 2016-10-07
Fix put_object_url and mime-type autodetect

faberge-eggs:
Couple bugfixies
put_object_url - add expires param
put_object - fix meme-type autodetect data.is_a?(File) returns false for file object


icco:
Nice catch! Thanks.

PR #167, opened 2016-10-21
Fix bug where listing of bucket causes infinite loop

### LINKED ISSUE #151: JSON Storage stack overflow when calling files.all()
Maybe I'm doing something wrong, but if I create a bucket and attempt to list files within, the system appears to get stuck in a loop until the stack is exhausted.

Steps to recreate (from HEAD):

```
irb(main):001:0> g = Fog::Storage::Google.new
=> [...]
irb(main):002:0> g.directories.create(key: 'example-fog-bucket').files
# repl hangs here, slowly exhausting stack
```

selmanj:
Closes #151 

Guys, I'm not sure that the json storage implementation is working; I couldn't find an implementation of `list_objects` and had to add one, and the file/directory abstractions contain a lot of stuff that doesn't match to the json api (hence the addition of the `name` alias change below in `file.rb`).

I'm trying to make this change as minor as possible in order to fix the bug, but I think it might be beneficial to revisit the entire storage json implementation.

### DIFF
--- a/lib/fog/storage/google_json/requests/list_objects.rb
+++ b/lib/fog/storage/google_json/requests/list_objects.rb
@@ -1,0 +1,50 @@
+module Fog
+  module Storage
+    class GoogleJSON
+      class Real
+        # Lists objects in a bucket matching some criteria.
+        #
+        # @param bucket [String] name of bucket to list
+        # @param options [Hash] optional hash of options
+        # @option options [String] :delimiter Delimiter to collapse objects
+        #   under to emulate a directory-like mode
+        # @option options [Integer] :max_results Maximum number of results to
+        #   retrieve
+        # @option options [String] :page_token Token to select a particular page
+        #   of results
+        # @option options [String] :prefix String that an object must begin with
+        #   in order to be returned
+        # @option options ["full", "noAcl"] :projection Set of properties to
+        #   return (defaults to "noAcl")
+        # @option options [Boolean] :versions If true, lists all versions of an
+        #   object as distinct results (defaults to False)
+        def list_objects(bucket, options = {})
+          api_method = @storage_json.objects.list
+          parameters = {
+            "bucket" => bucket
+          }
+
+          # Optional parameters
+          {
+            :delimiter   => "delimiter",
+            :max_results => "maxResults",
+            :page_token  => "pageToken",
+            :prefix      => "prefix",
+            :projection  => "projection",
+            :versions    => "versions"
+          }.each do |k, v|
+            parameters[v] = options[k] unless options[k].nil?
+          end
+
+          request(api_method, parameters)
+        end
+      end
+
+      class Mock
+        def list_objects(_bucket, _options = {})
+          Fog::Mock.not_implemented
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/fog/storage/google_json/requests/list_objects.rb
```diff
@@ -0,0 +1,50 @@
+module Fog
```
selmanj: Sorry, I'm not following you here, bot. Anyone know what he wants me to `freeze`?

erjohnso:
Looks like this was modeled after closely with fog-aws:
- Model matches: https://github.com/fog/fog-aws/blob/master/lib/fog/aws/models/storage/files.rb
- No list_objects request: https://github.com/fog/fog-aws/tree/master/lib/fog/aws/requests/storage

That said, as long is the functionality matches S3 (but works!), I think this is fine.

I don't know why your houndci is complaining. :(

Temikus:
IIRC there's also a need to be cross-compatible with Paperclip/CarrierWave. And AFAIK Paperclip works pretty well.

Thanks for the patch, I'll take this review, just give me a day or two, things are a bit hectic at the moment.

selmanj:
@Temikus Looking at both packages, it looks like they're both coded against the generic directories api. Paperclip doesn't give any examples on how to setup Google integration, but CarrierWave says to use S3 credentials via:

```ruby
CarrierWave.configure do |config|
  config.fog_provider = 'fog/google'                        # required
  config.fog_credentials = {
    provider:                         'Google',
    google_storage_access_key_id:     'xxxxxx',
    google_storage_secret_access_key: 'yyyyyy'
  }
  config.fog_directory = 'name_of_directory'
end
```

If someone has setup their fog client using the `google_storage_access_key_id`, then they will always use the XML interface for storage, not the JSON, as shown here: 
```ruby
module Fog
  module Storage
    class Google < Fog::Service
      def self.new(options = {})
        begin
          fog_creds = Fog.credentials
        rescue
          fog_creds = nil
        end

        if options.keys.include?(:google_storage_access_key_id) ||
           (!fog_creds.nil? && fog_creds.keys.include?(:google_storage_access_key_id))
          Fog::Storage::GoogleXML.new(options)
        else
          Fog::Storage::GoogleJSON.new(options)
        end
      end
    end
  end
end
```
Basically, if someone has setup their client to use storage credentials, then they're not using the JSON implementation (the XML implementation likely works fine). Also note this change isn't changing anything in the interface; instead I'm merely trying to make the interface work.

lib/fog/storage/google_json/requests/list_objects.rb
Temikus: @selmanj 
`# frozen_string_literal: true` is a magic comment, supported for the first time in Ruby 2.3, that tells Ruby that all string literals in the file are implicitly frozen. The comment must be on the first line of the file. It will be the default in Ruby3, so it's a way to test out the transition from 2.3 onwards.

I would avoid this for now as this needs to be added and tested en-masse.

Thanks for pointing this out, I'll disable the cop for now.

Temikus:
@selmanj I'm aware, I'm the one who reviewed that logic ;)

But AFAIK if one specifies the JSON credentials only, it should "just work"(TM) with Paperclip.
Not that sure about CarrierWave, I'll need to test that out.

And I get that you're only fixing the interface and I'm _extremely_ thankful for that. I want to just make sure we return the proper object from the get go, so we don't have to deprecate anything later.

Temikus:
LGTM, thank you so much Joe!

selmanj:
Awesome, thanks!

PR #168, opened 2016-10-22
Fix push_config to be proper structure

selmanj:
Existing code assigns the push_endpoint value directly to pushConfig,
which is not matching the structure defined at
https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions#PushConfig

Instead we just directly take the supplied hash.


Temikus:
LGTM. Thanks @selmanj!

PR #171, opened 2016-10-27
Add minitest pubsub tests, which test all the request methods

selmanj:
These use the minitest format rather than the shinto test-style. I tackled pubsub as I'm the most familiar with it, and I wanted to get early-feedback in case I'm writing minitests in a bad style.

### DIFF
--- a/test/integration/pubsub/test_pubsub.rb
+++ b/test/integration/pubsub/test_pubsub.rb
@@ -10,14 +10,21 @@ class TestPubsubService < FogIntegrationTest
 
   # Ensure any resources we create with test prefixes are removed
   Minitest.after_run do
-    @@client.list_topics[:body]["topics"].
-      map { |t| t["name"] }.
-      select { |t| t.start_with?(TOPIC_RESOURCE_PREFIX) }.
-      each { |t| @@client.delete_topic(t) }
-    @@client.list_subscriptions[:body]["subscriptions"].
-      map { |s| s["name"] }.
-      select { |s| s.start_with?(SUBSCRIPTION_RESOURCE_PREFIX) }.
-      each { |s| @@client.delete_subscription(s) }
+    topics = @@client.list_topics[:body]["topics"]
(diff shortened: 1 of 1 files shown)


test/integration/pubsub/test_pubsub.rb
```diff
@@ -0,0 +1,153 @@
+require "helpers/integration_test_helper"
+require "securerandom"
+require "base64"
+
+class TestPubsubService < FogIntegrationTest
+  @@client = Fog::Google::Pubsub.new
```
selmanj: I saw this error with rubocop and went back and forth on this.

I think the current code is easier to read, especially since it already has code that runs at class load time (specifically the minitest after_run code). I could factor it out to a class variable but given that the client is immutable (please correct me if i'm wrong) I'm not sure what gain we will get.

Also the above argument obviously only applies to test code; I wouldn't write this syntax in library code.

selmanj:
Also would appreciate if someone could run the pubsub test itself to ensure it works independently of my configuration.

test/integration/pubsub/test_pubsub.rb
icco: You don't really gain much here from a class variable, and they are considered bad style in ruby, so it'd be better to just create an instance variable.

selmanj:
@icco Made `@client` an instance variable now.

selmanj:
(Also added minor refactor of `some_subscription_name` and `some_topic_name`

Temikus:
Works like a charm! And even uses proper test credentials!
Thank you so much Joe!

```
temikus λ rake test --verbose                                                                                                                                            (1)
Run options: --seed 23987

# Running:

...........

Finished in 61.078333s, 0.1801 runs/s, 0.4912 assertions/s.

11 runs, 30 assertions, 0 failures, 0 errors, 0 skips
```

icco:
Thanks for this!

PR #176, opened 2016-11-08
Remove non-existant regions and add existing ones

selmanj:
Simple enough; just updates a failing test


Temikus:
LGTM thanks @selmanj

Issue #162, opened 2016-09-30
Fog::Compute::Google.get_target_pool_health doesn't allow you to specify an instance


selmanj:
The current implementation found in `fog-google/lib/fog/compute/google/requests/get_target_pool_health.rb` loops over every instance in the target_pool and makes an API call. Unfortunately the GCP api throws an error if the instance no longer exists or isn't created yet, which bubbles up to the caller.

This means that if a single instance doesn't exist in a target pool or isn't ready, then the caller has no way to identify the state of any of the other resources in that target_pool.  A simple fix is to just take the instance as an extra parameter.

Temikus:
Fixed by #177

PR #177, opened 2016-11-09
Add ability to check single instance health

Temikus:
WIP for #162 Do not merge yet.

@selmanj PTAL, did you have something like this in mind?

### DIFF
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -162,12 +162,12 @@ module Fog
           data.body["contents"]
         end
 
-        def set_disk_auto_delete(auto_delete, device_name=nil)
+        def set_disk_auto_delete(auto_delete, device_name = nil)
           requires :identity, :zone
 
           unless device_name
-            if self.disks.count <= 1
-              device_name = self.disks[0]['deviceName']
+            if disks.count <= 1
+              device_name = disks[0]["deviceName"]
             else
               raise ArgumentError.new("Device name required if multiple disks are attached")
             end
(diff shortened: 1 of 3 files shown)


selmanj:
Yeah, that'll do it (although it'd be nice if we could eliminate the conditional).

Speaking of, it looks like this takes a full instance url here; do we have any sort of convention on when to use the instance name vs the whole name within fog-google? I can see both sides (full url is clearer, but often redundant and annoying to type/find) and I'm curious what you think.

Temikus:
Well, the conditional will be there anyway, I should probably dedup the request logic though.
Taking a closer look at this - I think the logic should probably be moved to the model. Request really should just be a google-api-client proxy, requesting a single instance status. This will simplify a one-off call tremendously too, WDYT?

On the instance url, we prefer `instance_name`. However, I don't know how the API will behave in the future, so I don't want to guess the zone in this request.

selmanj:
Agreed on moving the logic to the model. You can make it backwards compatible by supporting the old behavior when instance name isn't provided (maybe with a deprecated message?)

Temikus:
@selmanj Cool, I'll start extracting.

I do not agree on deprecating the old behaviour though. I'd rather insert a rescue somewhere in the enumeration cycle. The thing is - Fog is designed to make cloud platform access easier. The current load balancer API is very difficult to work with, so we do some lifting on our side that the user can get more useful info.

selmanj:
Well, I was just advocating that you maintain the current behavior in the request method, but log a message saying the right place to satisfy this request is using the model class (since requests should just be simple request-based logic).  Then, sometime in the future you could remove the duplicated logic from the request code.

Temikus:
@selmanj PTAL

@icco - please, check whether the deprecation makes sense and if you agree with the returned data. I remember you used this.

PR #179, opened 2016-11-16
Add tests for addresses resource

selmanj:
This adds some basic end-to-end tests covering requests for the 'addresses' resource in compute. It's very similar to the pubsub tests.

I was going to use this as well as the pubsub tests to try to generalize some test helpers for testing basic CRUD testing without having to add all the boilerplate, but I'm holding off on that for now as my first attempt made the tests too hard to read. I'm opening a PR for this at the moment though as it's useful in its current form.

### DIFF
--- a/test/helpers/client_helper.rb
+++ b/test/helpers/client_helper.rb
@@ -1,0 +1,35 @@
+##
+# Helper mixin for tests (especially compute-based ones). Provide access to the
+# client in use via the #client method to use.
+
+module ClientHelper
+  # Check to see if an operation is finished.
+  #
+  # @param op [Excon::Response] the operation object returned from an api call
+  # @return [Boolean] true if the operation is no longer executing, false
+  #   otherwise
+  def operation_finished?(op)
+    # TODO: support both zone and region operations
+    region = op[:body]["region"]
+    name = op[:body]["name"]
+
+    result = client.get_region_operation(region, name)
+    !%w(PENDING RUNNING).include?(result[:body]["status"])
+  end
+
+  # Pause execution until an operation returned from a passed block is finished.
+  #
+  # @example Pause until server is provisioned
+  #   @result = wait_until_complete { client.insert_server(name, zone, opts) }
+  # @yieldreturn [Excon::Response] the resulting operation object from a block.
+  # @return [Excon::Response] the final completed operation object
+  def wait_until_complete
+    result = yield
+    return result unless result[:body]["kind"] == "compute#operation"
+
+    # TODO: support both zone and region operations
+    region = result[:body]["region"]
+    Fog.wait_for { operation_finished?(result) }
+    client.get_region_operation(region, result[:body]["name"])
+  end
+end
--- a/test/integration/compute/test_addresses.rb
+++ b/test/integration/compute/test_addresses.rb
@@ -1,9 +1,12 @@
 require "helpers/integration_test_helper"
+require "helpers/client_helper"
 require "securerandom"
 
 class TestComputeAddresses < FogIntegrationTest
   DEFAULT_REGION = "us-central1".freeze
   ADDRESS_RESOURCE_PREFIX = "fog-test-address".freeze
+
+  include ClientHelper
 
   # Ensure we clean up any created resources
   Minitest.after_run do
@@ -17,6 +20,8 @@ class TestComputeAddresses < FogIntegrationTest
     end
   end
 
+  attr_reader :client
+
   def setup
     @client = Fog::Compute::Google.new
   end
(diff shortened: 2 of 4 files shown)


test/helpers/client_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+# Helper mixin for tests (especially compute-based ones). Provide access to the
+# client in use via the #client method to use.
+
+module ClientHelper
+  # Check to see if an operation is finished.
+  #
+  # @param op [Excon::Response] the operation object returned from an api call
+  # @return [Boolean] true if the operation is no longer executing, false
+  #   otherwise
+  def operation_finished?(op)
```
Temikus: We already have a pretty powerful set of helpers for tracking operations:
```
[13] pry(main)> operation = server.set_tags({"world"=>"1"})
=>   <Fog::Compute::Google::Operation
...
 status="PENDING",
...
[14] pry(main)> operation.wait_for{ ready? }
=> {:duration=>0}
[15] pry(main)> operation.status
=> "DONE"
```

So are you sure this is needed? I may be misinterpreting this though.

test/integration/compute/test_addresses.rb
```diff
@@ -0,0 +1,84 @@
...
+class TestComputeAddresses < FogIntegrationTest
+  DEFAULT_REGION = "us-central1".freeze
+  ADDRESS_RESOURCE_PREFIX = "fog-test-address".freeze
+
+  include ClientHelper
+
+  # Ensure we clean up any created resources
+  Minitest.after_run do
+    client = Fog::Compute::Google.new
+    addresses = client.list_addresses(DEFAULT_REGION)[:body]["items"]
```
Temikus: As a rule of thumb - do not use raw api calls. This is why we have collections:
```
[17] pry(main)> gce.addresses
=> [  <Fog::Compute::Google::Address
    name="dns",
    kind="compute#address",
...
    status="RESERVED",
    users=nil
  >]
```

Temikus:
Ok, so, thanks for submitting this. However, some things will need to be fixed.

I ❤️  the style of the tests - they are very easy to read and follow. I see it being really easy to add new ones as well.

One major problem however - this is testing mainly requests, not models or collections, which should be the main unit of operation in fog.

To give you an idea:
Instead of `client.addresses`, you use raw call `client.list_addresses`, which tests mainly the request, not the model where all the logic resides.

Looking back at PubSub tests submitted earlier, they follow the same route. Sorry for not noticing this sooner - my bad 🙇

selmanj:
You're right; these are only testing the request methods and not the collection objects. My intention was to cover these with sanity tests so that, upon a change to the underlying google-api-client (like 0.9), we could have high confidence that fog-google isn't broken.

test/helpers/client_helper.rb
selmanj: I almost used the operation model, but I felt it was confusing because once you wrap a returned result with it, you no longer have access to the underlying excon result. To me, it was important to be able to assert low-level request properties on the request (like status code). It also tends to make the tests easier to understand (e.g. imagine tests on methods that return plain excon results mixed with tests that return operation model objects and assert on that).

If I weren't writing request-level tests though, I'd definitely be using the operation model here.

test/integration/compute/test_addresses.rb
selmanj: Again, I wanted to keep things on the raw api call level here since this test is about exactly that; it tends to make it easier to read.

Temikus:
@selmanj Ok, I see. And again, thanks for tackling this. However, I do have 2 big concerns:

1) IMO one of the concepts of Ruby is that tests should follow the actual usage of the code, a lot of people look at tests to figure out how something should be used, with this we're showing people a pattern that should not be used at all (using raw requests instead of models).
2) Those test google-api-client and requests more, rather than what fog was actually created for. I do realise that those are easier to read and make it easier to transition to new google-api-client and check for api changes, however, if the transition breaks any of the actual logic - we're still in the dark, and added a 3rd set of tests that sorta covers one of the parts of the lib :|

With those in mind - do you plan to add model/collection tests separately and clearly separate them in tests or are those not on the roadmap?

selmanj:
> IMO one of the concepts of Ruby is that tests should follow the actual usage of the code, a lot of people look at tests to figure out how something should be used, with this we're showing people a pattern that should not be used at all (using raw requests instead of models).

I wonder why raw requests shouldn't be used at all?  The [general fog page](http://fog.io/about/structure.html) specifically calls them out as something to be used when models don't cut it. I don't think all the requests are necessarily covered by models, are they?

> Those test google-api-client and requests more, rather than what fog was actually created for. I do realise that those are easier to read and make it easier to transition to new google-api-client and check for api changes, however, if the transition breaks any of the actual logic - we're still in the dark, and added a 3rd set of tests that sorta covers one of the parts of the lib :|

I'm trying to be very careful *not* to test google-api-client and instead test the functionality of the request methods themselves; instead I am testing that the low-level request object returns some reasonable representation. The hope is that if the tests are passing, then all of the models "just work".

I totally understand being hesitant about more tests that aren't complete, but I tried to write them so that they're independent of each other, and more importantly they pass consistently.

> With those in mind - do you plan to add model/collection tests separately and clearly separate them in tests or are those not on the roadmap?

That's correct; in my mind the request-based method were higher-priority since they are blocking updating the underlying google-api-client.  Does this sound reasonable?

selmanj:
I added `test_compute_addresses_collection.rb` which tests the basics of the addresses collection, while also showing a walkthrough of how to use the api. Let me know what you think.

Temikus:
Played around with those a bit. Looks solid. Thanks!

Nit, folder structure already speaks for itself, so:
- test_compute_address_requests should probably just be `test_address_requests`
- test_compute_addresses_collection should be `test_addresses`

test/helpers/client_helper.rb
Temikus: ACK, LGTM

selmanj:
@Temikus The redundant naming is because the tests don't have any sort of namespacing, and it wasn't clear to me if MiniTest loads all the tests at once and then runs or one at a time.  For example, if I have two `TestNameRequests` classes in different paths, will minitest flip out when loading/running them?

I wonder if it would be clearer if the tests were namespaced (e.g. `IntegrationTests::Compute::TestRequests`)

Temikus:
@selmanj Good point. Ack, merging. Thank you very much for this.

PR #186, opened 2016-12-26
Make server.add_ssh_key an operation

Temikus:
Follow-up to #185 

It seems that add_ssh_key never actually performed an operation but rather changed the metadata of the currently saved `server` object. Since one can ?no longer? call `server.save` on an existing instance, this method seems useless without an actual call to metadata update.

@icco @selmanj - PTAL, since I think I may be missing something.

P.S. Updated all the instance templates, since some older wheezy images do not support `ssh-key` metadata + we had some images from !2014! in tests/bootstrap.

### DIFF
--- a/examples/create_instance.rb
+++ b/examples/create_instance.rb
@@ -9,13 +9,12 @@ Bundler.require(:default, :development)
 def test
   connection = Fog::Compute.new(:provider => "Google")
 
-  name = "fog-smoke-test-#{Time.now.to_i}"
-
   disk = connection.disks.create(
-    :name => name,
+    :name => "fog-smoke-test-#{Time.now.to_i}",
     :size_gb => 10,
     :zone_name => "us-central1-f",
-    :source_image => "debian-7-wheezy-v20151104")
+    :source_image => "debian-8-jessie-v20161215"
+  )
 
   disk.wait_for { disk.ready? }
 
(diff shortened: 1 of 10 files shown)


icco:
So I have no idea why your image tests are failing. That's a totally valid build image.

Temikus:
Yeah, this is because I forgot to update `mock.rb`.

Looking at it - it's ancient, some of the fields in images do not correspond to the real thing anymore it seems. I updated both the model and the mocks by hand.

@icco - do we have any way to update those automatically, or at least, spew something out I can readily copy/paste in? Otherwise it's quite a bit tedious, even if I take raw output from `gcloud images list --format json`

selmanj:
Hmm; is it surprising that add_ssh_key performs an operation now? What if someone is doing a for-loop adding a large set of ssh keys - this could result in a noticeable delay now.

Would it be possible to change `#save` to instead check to see if the instance already exists, and just update the metadata if that's all that's changed? That would be more in line with the fog-style api IMO

Temikus:
@selmanj thanks for noticing this. On the question of large amount of keys - if this is a use case - it should be done with bulk metadata update. This method should IMO be a convenience method for the server model.

What is surprising is that something that can only be modified during the instance creation is a public method :/ 

I wouldn't overload 'save' with logic - there's a bunch of immutable characteristics on Google side already and making a subset of them mutable seems like a bad idea to me. Not even mentioning the fact that core model manipulation is already quite finicky.

PR #187, opened 2016-12-26
Update .travis.yml

icco:
Add ruby 2.4 and turn on caching to speed up builds.


Temikus:
Thanks Nat!

We'll need to figure out how to upgrade json, seems that's why the build is failing on 2.4

PR #193, opened 2017-01-10
Pass google client options to the correct client

selmanj:
Also reformatted up the yard doc while documenting the additional parameter


Temikus:
Merging, thanks @selmanj

PR #198, opened 2017-01-12
Add recognize for google client options

### LINKED ISSUE #197: google_client_options is an unrecognized option
The new `:google_client_options` exposed via the constructor is not listed as a recognized option, which leads to a warning (despite it actually being recognized). Luckily it's still used, but it's a bit noisy.

selmanj:
Cleaned up the recognizes option a bit and added :google_client_options to suppress warning.

Fixes #197

### DIFF
--- a/lib/fog/google/pubsub.rb
+++ b/lib/fog/google/pubsub.rb
@@ -10,6 +10,7 @@ module Fog
         :google_key_location,
         :google_key_string,
         :google_client,
+        :google_client_options,
         :app_name,
         :app_version,
         :google_json_key_location,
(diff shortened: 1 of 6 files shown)


icco:
Could you sort all of the recognizes lists so they are similar?

selmanj:
@icco Sure, done.

lib/fog/google/pubsub.rb
```diff
@@ -5,8 +5,17 @@ class Pubsub < Fog::Service
...
-                 :app_name, :app_version, :google_json_key_location, :google_json_key_string
+      recognizes(
+        :google_client_email,
+        :google_key_location,
+        :google_key_string,
+        :google_client,
+        :google_client_options,
+        :app_name,
+        :app_version,
+        :google_json_key_location,
```
icco: This ones out of order

selmanj: oh geez, how did i miss that

PR #199, opened 2017-01-13
Releasing 0.5.2

Temikus:
Quick releasing 0.5.2 due to regression in #190 still present in 0.5.1


Temikus:
@selmanj Fixed the regression, pushed the release.

We should get better at release notes too.

PR #203, opened 2017-02-08
Allow creation of an instance group in a network other than 'default'

DimitriosLisenko:


lib/fog/compute/google/models/instance_group.rb
```diff
@@ -60,6 +64,10 @@ def zone_name
           zone.nil? ? nil : zone.split("/")[-1]
         end
 
+        def network_name
+          network.nil? ? nil : network.split("/")[-1]
```
DimitriosLisenko: To me, since this is a method on the `InstanceGroup` object, it feels weird to return a value for something absent in the object itself. It seems more correct to assume a default value if it still hasn't been supplied at the point of creation in `insert_instance_group`.

icco: Seems reasonable.

PR #212, opened 2017-04-14
IsTruncated doesn't work in json storage api

### LINKED ISSUE #206: JSON-API: `is_truncated` missing in Files
Hello,

last week, when attempting to do some `all(prefix: "...") { ... }` operation on a `Fog::Storage::GoogleJSON::Files` collection of a given directory, to my surprise, I received a `NoMethodError` for `is_truncated`.

In [v0.5.2 of this file, line 31](https://github.com/fog/fog-google/blob/v0.5.2/lib/fog/storage/google_json/models/files.rb#L31) (and older ones), there is a check for `is_truncated`. Looking at the rest of the code here, there is no such thing - but when comparing it to the[ XML version](https://github.com/fog/fog-google/blob/v0.5.2/lib/fog/storage/google_xml/models/files.rb#L11), there is this attribute.

Unfortunately, I cannot promise to look deeper into the JSON protocol to set things correctly, and we took another approach to work around this. I tried injecting the attribute from outside once which eradicated the error but seemed to ignore my `prefix` setting so far, giving me the wrong items.

Please keep track of this error. Thank you.

icco:
This should fix #206. @MarcelHB, could you verify this fixes your bug? @Temikus mind giving a quick review?

### DIFF
--- a/lib/fog/storage/google_json/models/files.rb
+++ b/lib/fog/storage/google_json/models/files.rb
@@ -28,10 +28,6 @@ module Fog
             subset = dup.all
 
             subset.each_file_this_page { |f| yield f }
-            while subset.is_truncated
-              subset = subset.all(:marker => subset.last.key)
-              subset.each_file_this_page { |f| yield f }
-            end
 
             self
           end
--- a/lib/fog/storage/google_json/requests/get_bucket.rb
+++ b/lib/fog/storage/google_json/requests/get_bucket.rb
@@ -61,13 +61,11 @@ module Fog
               response.body = {
                 "CommonPrefixes"  => [],
                 "Contents"        => truncated_contents,
-                "IsTruncated"     => truncated_contents.size != contents.size,
                 "Marker"          => options["marker"],
                 "Name"            => bucket["Name"],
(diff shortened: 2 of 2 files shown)


MarcelHB:
Sorry, I no longer have access to a Google storage env. Someone else required to test this. :man_shrugging:

Temikus:
Ok, so - LGTM on the change itself, since this is a no longer supported parameter anyway.
We probably should prune some others, like marker outta there -_-

However, it seems that either I'm misunderstanding the usage or `prefix` is indeed not working:
```
pry(main)> storage.directories.get("fog-testing").files.all(prefix: "02").first
=>   <Fog::Storage::GoogleJSON::File
    key="02059_bridge_1920x1080.jpeg",
    acl=nil,
    predefined_acl=nil,
```
Should I spend some time debugging this?

Temikus:
@icco NVM, overlooked that it's in the enumerator. Reproduced. Sorry, my head's not working today it seems.

Repro:
```
[6] pry(main)> storage.directories.get("fog-testing").files.all(prefix: "02").each {|f| f.acl} ;
[7] pry(main)>
```

PR #213, opened 2017-04-19
add support for update_firewall

cowboyrushforth:
this commit allows one to update the firewall rules

### DIFF
--- a/lib/fog/compute/google/models/firewall.rb
+++ b/lib/fog/compute/google/models/firewall.rb
@@ -28,6 +28,15 @@ module Fog
           reload
         end
 
+	def update
+          requires :identity, :allowed, :network
+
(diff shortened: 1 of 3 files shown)


lib/fog/compute/google/models/firewall.rb
```diff
@@ -28,6 +28,15 @@ def save
           reload
         end
 
+        def update
```
icco: So instead of calling update, could we make this be logic in `save`? So save will insert if it doesn't exist or update if it does?

cowboyrushforth: I like the idea of this, but dislike the idea of having every call to save having to first make a request to see if it should do an insert or save.  Is that what you are suggesting, or is there a different way you were envisioning?  Thanks!

Issue #214, opened 2017-04-24
subscription push_config does not default to empty hash


durandom:
I'm getting 
```
[NoMethodError]: undefined method `empty?' for nil:NilClass
/opt/rh/cfme-gemset/gems/fog-google-0.5.2/lib/fog/google/requests/pubsub/create_subscription.rb:29:in `create_subscription'
/opt/rh/cfme-gemset/gems/fog-google-0.5.2/lib/fog/google/models/pubsub/subscription.rb:71:in `save'
/opt/rh/cfme-gemset/gems/fog-core-1.43.0/lib/fog/core/collection.rb:51:in `create'
```

The calling code is 
```ruby
google.subscriptions.create(:name  => subscription_name, :topic => topic_name)
```

My guess is either https://github.com/fog/fog-google/pull/168 introduced this regression because [this line](https://github.com/fog/fog-google/blob/4ac58a77f91f4d6877d77133b6bcee33403951d8/lib/fog/google/models/pubsub/subscription.rb#L71) passes `push_config` and if it's `nil` it will pass `nil`.

Adding `push_config => {}` to the caller fixes the issue.

PR #215, opened 2017-04-24
check if push_config is present before checking for entries

durandom:
I'm getting 
```
[NoMethodError]: undefined method `empty?' for nil:NilClass
/opt/rh/cfme-gemset/gems/fog-google-0.5.2/lib/fog/google/requests/pubsub/create_subscription.rb:29:in `create_subscription'
/opt/rh/cfme-gemset/gems/fog-google-0.5.2/lib/fog/google/models/pubsub/subscription.rb:71:in `save'
/opt/rh/cfme-gemset/gems/fog-core-1.43.0/lib/fog/core/collection.rb:51:in `create'
```

The calling code is 
```ruby
google.subscriptions.create(:name  => subscription_name, :topic => topic_name)
```

My guess is either https://github.com/fog/fog-google/pull/168 introduced this regression because [this line](https://github.com/fog/fog-google/blob/4ac58a77f91f4d6877d77133b6bcee33403951d8/lib/fog/google/models/pubsub/subscription.rb#L71) passes `push_config` and if it's `nil` it will pass `nil`.

Adding `push_config => {}` to the caller fixes the issue.

this fixes https://github.com/fog/fog-google/issues/214

I'm not sure how and where to add a test for this. Seems this code path is not tested / covered at all :(
If you point me to a similar test, I can add it

### DIFF
--- a/lib/fog/google/requests/pubsub/create_subscription.rb
+++ b/lib/fog/google/requests/pubsub/create_subscription.rb
@@ -26,7 +26,7 @@ module Fog
             "topic" => (topic.is_a?(Topic) ? topic.name : topic.to_s)
           }
 
-          unless push_config.empty?
+          if push_config && push_config.any?
             body["pushConfig"] = push_config
           end
 


durandom:
cc @selmanj maybe you can have a quick look, as this is some of your code ? thanks...

lib/fog/google/requests/pubsub/create_subscription.rb
```diff
@@ -26,7 +26,7 @@ def create_subscription(subscription_name, topic, push_config = {}, ack_deadline
             "topic" => (topic.is_a?(Topic) ? topic.name : topic.to_s)
           }
 
-          unless push_config.empty?
+          if push_config && push_config.any?
```
icco: Had never used `.any?` before, but TIL!

```
2.4.0 :002 > nil && nil.any?
 => nil
2.4.0 :003 > {} && {}.any?
 => false
2.4.0 :004 > {a: 'b'} && {a: 'b'}.any?
 => true
```

Does not look like `tests/models/pubsub/` has any tests for this, nor does `tests/requests`. So I guess it's fine to just not have any for now.

icco:
Thanks for contributing!

Temikus:
@durandom - there should be integration tests for this in https://github.com/fog/fog-google/blob/master/test/integration/pubsub/test_pubsub_requests.rb

If those do not cover it - let me know and I'll see if I can add some.

Thanks for contributing!

durandom:
@Temikus I just added a byebug breakpoint in the codepath, like [here](https://github.com/durandom/fog-google/blob/f7da96c9df1b15f49aa88a57f70ba2434fd95f1a/lib/fog/google/requests/pubsub/create_subscription.rb#L20) and ran the tests and it was not triggered. So I assume that path is not covered....

PR #222, opened 2017-05-18
Fixes #221 - Use nextPageToken to list all snapshots

### LINKED ISSUE #221: Listing snapshot should respect nextPageToken so more than 500 snapshot could be listed
Currently nextPageToken is ignored, resulting a limit of 500 snapshots.
See https://cloud.google.com/compute/docs/reference/latest/snapshots/list for more details.

imriz:


icco:
Seems straight forward! Thanks @imriz!

PR #227, opened 2017-06-26
Upgrade PubSub requests to >= 0.11

Everlag:
This breaks compatibility with the previous excon based approach
as keeping it would require us to manually wrap every API
client operation in an excon response.
Additionally, all mocks have been deleted and left unimplemented.

However, the tests have been migrated and currently pass.

Looking for review to make sure this style is fine to continue with for the pubsub models. Thanks!

### DIFF
--- a/test/integration/pubsub/test_pubsub_requests.rb
+++ b/test/integration/pubsub/test_pubsub_requests.rb
@@ -12,20 +12,34 @@ class TestPubsubRequests < FogIntegrationTest
   end
 
   def delete_test_resources
-    topics = @client.list_topics[:body]["topics"]
-    unless topics.nil?
-      topics.
-        map { |t| t["name"] }.
-        select { |t| t.start_with?(topic_resource_prefix) }.
-        each { |t| @client.delete_topic(t) }
+    topics_result = @client.list_topics
+    unless topics_result.topics.nil?
+      begin
+        topics_result.topics.
+          map(&:name).
+          select { |t| t.start_with?(topic_resource_prefix) }.
+          each { |t| @client.delete_topic(t) }
+      # We ignore errors here as list operations may not represent changes applied recently.
+      # Hence, list operations can return a topic which has already been deleted but which we
+      # will attempt to delete again.
+      rescue Google::Apis::Error
+        Fog::Logger.warning("ignoring Google Api error during delete_test_resources")
+      end
     end
 
-    subscriptions = @client.list_subscriptions[:body]["subscriptions"]
-    unless subscriptions.nil?
-      subscriptions.
-        map { |s| s["name"] }.
-        select { |s| s.start_with?(subscription_resource_prefix) }.
-        each { |s| @client.delete_subscription(s) }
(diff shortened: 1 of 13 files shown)


Everlag:
@icco @Temikus PubSub model and storage PRs are blocked on this, comments here would be appreciated as those effect pipelined changes.

icco:
Totally, reviewing now.

test/integration/pubsub/test_pubsub_requests.rb
```diff
@@ -12,20 +12,32 @@ def setup
...
+    topics_result = @client.list_topics
+    unless topics_result.topics.nil?
+      begin
+        topics_result.topics.
+          map(&:name).
+          select { |t| t.start_with?(topic_resource_prefix) }.
+          each { |t| @client.delete_topic(t) }
+      # We ignore errors here as they are flaky due to the delay in list operations
+      # representing our operations.
+      rescue Google::Apis::Error
```
icco: Are there not specific errors relating to just listing? If not that's fine, it'd just be cool if there was.

test/integration/pubsub/test_pubsub_requests.rb
```diff
@@ -12,20 +12,32 @@ def setup
...
+        puts "ignoring Google Api error during delete_test_resources"
```
icco: Fog has a logging interface, please use that over puts.

test/integration/pubsub/test_pubsub_requests.rb
```diff
@@ -50,117 +62,123 @@ def some_topic_name
...
     end
-    @some_subscription
   end
 
-  def test_create_topic
-    result = @client.create_topic(new_topic_name)
+  # retry_times attempts the given block up to tries times.
+  # A failed attempt to considered to be any point where the block yields
+  # and a Google::Apis::Error or StandardError is rescued.
+  def retry_times(tries, &block)
```
icco: If possible, it'd be nice to use https://github.com/fog/fog-core/blob/master/lib/fog/core/wait_for.rb instead of creating a oneoff function

test/integration/pubsub/test_pubsub_requests.rb
Everlag: Clarified, the issue isn't getting the list, its that the list potentially has already deleted entries.

test/integration/pubsub/test_pubsub_requests.rb
Everlag: Done

test/integration/pubsub/test_pubsub_requests.rb
Everlag: Done, thanks for showing me that.

icco:
LGTM. I can merge whenever you're ready @Everlag

Everlag:
Great. I'd prefer to wait on merging until the models are also ready. PR for models should be sometime later today.

Everlag:
Actually, can we merge now? Sorry about that, I'm used to gerrit's method of handling changes, that model falls apart on github.

PR #218, opened 2017-05-15
Cleaning up tests

Temikus:
Anyone working on this project should be able to just run tests with
`rake tests` without figuring out our testing legacy, otherwise I'm
uncomfortable making people run tests.

- Deactivating nonfunctioning tests to be either deleted or reworked
- Making sure the tests actually do their thing in default config

Work towards #51 

Now it actually works (yes, the 2 failing tests are new, I'll get on it):
```
2.1.7 ~/Code/fog-devel/fog-google  (test_workflow*)-temikus@google.com [16:37:06]∞
temikus λ rake test                                                                                                                                                     (1)
Run options: --seed 15446

# Running:

............EE......

Finished in 137.918791s, 0.1450 runs/s, 0.4495 assertions/s.

  1) Error:
UnitTestServer#test_add_ssh_key_uses_ssh_keys_by_default:
ArgumentError: service, identity and zone are required for this operation
    /Users/temikus/.rbenv/versions/2.1.7/lib/ruby/gems/2.1.0/gems/fog-core-1.43.0/lib/fog/core/attributes.rb:157:in `requires'
    /Users/temikus/Code/fog-devel/fog-google/lib/fog/compute/google/models/server.rb:188:in `set_metadata'
    /Users/temikus/Code/fog-devel/fog-google/lib/fog/compute/google/models/server.rb:214:in `add_ssh_key'
    /Users/temikus/Code/fog-devel/fog-google/test/unit/compute/test_server.rb:15:in `test_add_ssh_key_uses_ssh_keys_by_default'


  2) Error:
UnitTestServer#test_metadata_uses_deprecated_sshKeys_if_exists:
ArgumentError: service, identity and zone are required for this operation
    /Users/temikus/.rbenv/versions/2.1.7/lib/ruby/gems/2.1.0/gems/fog-core-1.43.0/lib/fog/core/attributes.rb:157:in `requires'
    /Users/temikus/Code/fog-devel/fog-google/lib/fog/compute/google/models/server.rb:188:in `set_metadata'
    /Users/temikus/Code/fog-devel/fog-google/lib/fog/compute/google/models/server.rb:214:in `add_ssh_key'
    /Users/temikus/Code/fog-devel/fog-google/test/unit/compute/test_server.rb:7:in `test_metadata_uses_deprecated_sshKeys_if_exists'

20 runs, 62 assertions, 0 failures, 2 errors, 0 skips


rake aborted!
[...]

PR #229, opened 2017-07-07
Remove failing compute and storage test suites

Everlag:
These tests are written against the old api client, hence they
will always fail as we are no longer providing excon responses.
These will be rewritten as their corresponding functionality is
reimplemented.

Having these gone now should make for cleaner diffs on storage
and compute tests.

This has the same effect as #218 did but favors removal rather
than carrying inactive tests forward.


Everlag:
Can you point out the `collection` tests which should be passing? I'm having some trouble finding them and would rather not delete working examples.

Temikus:
Those are collection tests (ending with `_collection`):
test_compute_addresses_collection.rb
test_compute_networks_collection.rb
, they should be request-independent. If something's broken in them - we most likely have broken something.

Temikus:
Also, while we're at it - did you see the Travis errors about no `managed_zones` method?

Everlag:
I agree, the collection tests should eventually pass, I'll remove them from this change.

For the managed_zones error, I've only been considering the minitest suites, would you prefer the shindo tests also be handled here?

Everlag:
I've updated this PR to not conflict with the half-merged Storage changes, #231 and #233. I think this is worth looking at to get useful results out of concourse while we migrate the api client.

icco:
Just going to close for now, may revisit later.

PR #210, opened 2017-04-07
Upgrade google client to 0.15

icco:
This is very much ongoing, but I'd love a first pass whenever people get a chance. Using this branch because #100 never really got going.

Work is being tracked in #226

PR #230, opened 2017-07-13
Issue with signed urls and paths containing white spaces

andresbravog:
# What

We have experienced while using `paperclip` + `fog` + `fog-google` that files containing special characters (white space) where incorrectly translated to `+` char. This is because of CGI.escape usage instead of URI.encode.

CGI.scape does translate ` ` to `+` as in `https://` . And `+` means a space only in `application/x-www-form-urlencoded` content, such as the query parameter keys.

As we are forming urls as so:

```
https://storage.googleapis.com/MY_BUCKET/RANDOM NAME-with_chars.jpgGoogleAccessId=REDACTED&Signature=REDACTED&Expires=3600
```

this will endup being transformed to

```
https://storage.googleapis.com/MY_BUCKET/RANDOM+NAME-with_chars.jpgGoogleAccessId=REDACTED&Signature=REDACTED&Expires=3600
```

ending in `NoSuchKey` error. Instead of

```
https://storage.googleapis.com/MY_BUCKET/RANDOM%20NAME-with_chars.jpgGoogleAccessId=REDACTED&Signature=REDACTED&Expires=3600
```

what is the correct one.

### DIFF
--- a/lib/fog/storage/google_json/utils.rb
+++ b/lib/fog/storage/google_json/utils.rb
@@ -1,3 +1,5 @@
+require 'uri'
+
 module Fog
   module Storage
     class GoogleJSON
@@ -19,7 +21,7 @@ module Fog
 
         def host_path_query(params, expires)
           params[:headers]["Date"] = expires.to_i
-          params[:path] = CGI.escape(params[:path]).gsub("%2F", "/")
+          params[:path] = URI.encode(params[:path]).gsub("%2F", "/")
           query = [params[:query]].compact
           query << "GoogleAccessId=#{@client.authorization.issuer}"
           query << "Signature=#{CGI.escape(signature(params))}"
--- a/lib/fog/storage/google_xml/utils.rb
+++ b/lib/fog/storage/google_xml/utils.rb
@@ -1,3 +1,5 @@
+require 'uri'
+
 module Fog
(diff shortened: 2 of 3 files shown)


icco:
So, both work, and https://cloud.google.com/storage/docs/naming#object-considerations doesn't really say. For some reason, I remember there being a reason for us using CGI escaping instead of URI escaping at the time of initial implementation... but that could also be just an incorrect memory.

@andresbravog for now, I'd suggest doing the escaping ahead of passing into Fog. As we are currently refactoring all code in #210, and will be doing a major release (and this will make files for existing users unaccessible via fog if they don't know about this change. 

@Temikus thoughts? @Everlag could you make sure this change makes it into your refactor?

Everlag:
I'll include the json change in the PR for models.

andresbravog:
Cool! thanks @Everlag and @icco. Feel free to close this PR if you guys are gonna handle it in some other way.

cheers!

Temikus:
Sorry, I've been out of it for quite a while. For posterity's sake:

We actually should use special escape, implemented to work with all corner-cases:
https://github.com/fog/fog-google/commit/c7b12483ecef82af0fbfb690b562dc56aa201d8d

I'll create an issue.

PR #233, opened 2017-07-17
Upgrade Storage models to >= 0.11

Everlag:
This includes the models along with some Storage request corrections, test fixes, and the fix for #230


lib/fog/storage/google_json/models/directory.rb
```diff
@@ -29,31 +21,18 @@ def files
...
         def public_url
           requires :key
-          acl = service.get_bucket_acl(key).body
-          if acl["items"].detect { |entry| entry["entity"] == "allUsers" && entry["role"] == "READER" }
-            if key.to_s =~ /^(?:[a-z]|\d(?!\d{0,2}(?:\.\d{1,3}){3}$))(?:[a-z0-9]|\.(?![\.\-])|\-(?![\.])){1,61}[a-z0-9]$/
-              "https://#{key}.storage.googleapis.com"
-            else
-              "https://storage.googleapis.com/#{key}"
-            end
+          if key.to_s =~ /^(?:[a-z]|\d(?!\d{0,2}(?:\.\d{1,3}){3}$))(?:[a-z0-9]|\.(?![\.\-])|\-(?![\.])){1,61}[a-z0-9]$/
```
icco: What is this regex?

Everlag:
ACL changes are detailed in the 'Upgrade Storage models to >= 0.11' commit message:

Semantics of public_url have been changed. We no longer
validate the ACL as it is impossible to mutate
those ACLs as a result of the below changes. Hence,
we simply return the url which would provide access to
the resource if it was public.

Requests and their tests have been modify to validate
creating objects and buckets with predefined ACLs.

The ACL mutations, which operated mainly using predefined
ACLs, have been removed as they were previously unimplemented.
When I attempted to implement them, I ran into a semantics
mismatch with what we are allowed to use predefined ACLs for.
You can create an object or bucket with a predefined ACL.
You cannot update an object or bucket with a predefined ACL.
I confirmed this behavior is acknowledged internally
and will not be changed.

This was mentioned in google-cloud-ruby
GoogleCloudPlatform/google-cloud-ruby#381
(Final paragraph)

To reproduce this ACL issue:
Create a new bucket in GCP console, add an object using upload.
Try to use the 'Objects: patch' method with the API Explorer
to change to a predefined ACL. You will encounter the following error

```
"code": 409,
  "message": "Cannot provide both a predefinedAcl and access controls."
```

https://cloud.google.com/storage/docs/json_api/v1/objects/patch

lib/fog/storage/google_json/models/directory.rb
Everlag: I'm honestly not sure, the change here was me factoring out the ACL validation.

icco:
Wow, https://cloud.google.com/storage/docs/json_api/v1/objects/get says
nothing about this.

https://cloud.google.com/storage/docs/xml-api/reference-uris is where this
comes from, but I think we should remove the regex and pick a format. I
personally prefer the bucket in the path, but maybe ask around which is
preferred... Or do what the console does?

Everlag:
Updated to always include bucket in the path, that follows the GCP console convention.

Everlag:
:+1: 

In the future, we could expose manipulating non-canned ACLs. Probably a post-migration problem, want me to open a tracking issue?

icco:
@Everlag that'd be great.

PR #235, opened 2017-07-19
Upgrade Compute addresses to >= 0.11

Everlag:
Also upgraded are the operation models and requests,
minimally to get the tests working.

This is a shift from my previous strategy of PRing requests
first then models for an entire module. That does not scale to
compute. For Compute, I'll PRing specific models as which will be
as implemented as possible, changing which requests are necessary.

This commit also proposes a change in the Compute test layout.
A flat directory, as tests are currently laid out in, makes it
difficult to match associated request and model tests.
This new layout introduces a directory per-model.

Note: there are several unimplemented portions of Address.
These are dependent on the server implementation and will land
when possible.


lib/fog/compute/google/models/address.rb
```diff
@@ -45,7 +47,10 @@ def destroy(async = true)
           requires :identity, :region
 
           data = service.delete_address(identity, region.split("/")[-1])
-          operation = Fog::Compute::Google::Operations.new(:service => service).get(data.body["name"], nil, data.body["region"])
+          operation = Fog::Compute::Google::Operations.
+                      new(:service => service).
```
icco: TIL we do trailing dots... ew. Oh well.

lib/fog/compute/google/models/operations.rb
```diff
@@ -5,26 +5,27 @@ class Operations < Fog::Collection
...
         def all(filters = {})
-          if filters["zone"]
-            data = service.list_zone_operations(filters["zone"]).body
-          elsif filters["region"]
-            data = service.list_region_operations(filters["region"]).body
-          else
-            data = service.list_global_operations.body
-          end
-          load(data["items"] || [])
+          data = if filters.zone
```
icco: Oh god, I thought we removed this style...

lib/fog/compute/google/models/address.rb
Everlag: Yeah, there was a double take when rubocop told me to do that :|

lib/fog/compute/google/models/operations.rb
Everlag: Given that we're rewriting compute and it's >9k lines, this would be a good time to do that.

Everlag:
Concourse failed with an unexpected error, I'm looking into it and will :+1: when fixed.

Thanks for the review.

Everlag:
Fixed the issue, :+1: to merge.

PR #241, opened 2017-08-12
Upgrade google client monitoring

DawidJanczak:
Some of the old integration tests are failing, what should I do about them? I noticed we generally `raise Fog::Errors::MockNotImplemented` inside the respective mocks, should I do the same here?


icco:
Yeah, we haven't figured out how to deal with mocks in the new api world, feel free to replace with raise for now. I'll add the integrate label which will run the tests with real infra.

DawidJanczak:
@icco Got it, done.

PR #242, opened 2017-08-14
Remove broken mocks from DNS for Google Client 0.13

icco:


icco:
Ok, cleaned out all of the broken tests, now going to migrate them to minitest I guess. Faster would probably be to just override shindo to not error on mock not implemented exceptions, but I dislike this mixed test world we are in.

icco:
Going to merge this and then go forward.

PR #246, opened 2017-09-03
Upgrade google client backend services

DawidJanczak:
@icco I'll need your help with the tests. This https://github.com/fog/fog-google/blob/master/test/integration/factories/backend_services_factory.rb#L17 seems to be failing with `NoMethodError: undefined method `http_health_checks' for #<Google::Apis::ComputeV1::ComputeService:0x0055b5c85dbc80>` Should I just remove this param and the http health checks factory?

### DIFF
--- a/lib/fog/compute/google/models/backend_service.rb
+++ b/lib/fog/compute/google/models/backend_service.rb
@@ -30,7 +30,7 @@ module Fog
           }
 
           data = service.insert_backend_service(name, options).body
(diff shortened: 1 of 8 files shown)


icco:
I think so, that makes sense @DawidJanczak .

icco:
@DawidJanczak I don't think I totally understand why the test is breaking, but are healthchecks implemented yet? It may be they need to be ported over at the same time.

Also, sorry I took so long on this.

icco:
I'm going to merge this and we can sort out the tests later.

PR #250, opened 2017-10-05
Migrate upgrade-google-client monitoring to v3

emilymye:
I still need to support groups before we are feature complete, but I added some tests. The tests will probably incur stackdriver monitoring costs and we can't delete timeseries :( Please let me know if this is/becomes an expensive test run.


emilymye:
@selmanj please take a look as well if you have time?

icco:
This looks great, thanks!

PR #252, opened 2017-10-18
Part 1 of Cloud SQL upgrade

emilymye:
Part 1: 
- Update SQL client to use new google api client.
- Updated: flags, tiers

Resources TODO (all involving instances, which is the largest chunk):
- instances
- operations
- ssl_certs
- backup_runs


emilymye:
sgtm - I assume it's ok that some of the other sql stuff just won't work right now? I'll go ahead and get the next PR out today or tomorrow.

icco:
Yup!

PR #253, opened 2017-10-21
Upgrade google client instance groups

DawidJanczak:
A few questions/things I'm not sure of:

1) I haven't reimplemented the deprecated methods here https://github.com/fog/fog-google/blob/master/lib/fog/compute/google/models/instance_groups.rb#L35-L51 - should I? FWIW I'm not sure they even work, e.g. `unless params[:instance] == Array` is missing one `=` to make sense :)
2) `InstanceGroup#save` throws `alreadyExists` error - is that normal? I can change the name and attempt to save it, but then it fails on `subnetwork` as that's apparently ready-only.

### DIFF
--- a/lib/fog/compute/google/models/instance_group.rb
+++ b/lib/fog/compute/google/models/instance_group.rb
@@ -53,10 +53,10 @@ module Fog
           requires :identity, :zone
 
           instance_list = []
-          data = service.list_instance_group_instances(identity, zone_name).body
-          if data["items"]
-            data["items"].each do |instance|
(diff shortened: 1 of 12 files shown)


icco:
1. Yeah don't implement the deprecated stuff.
2. I'm not sure based off of https://cloud.google.com/compute/docs/reference/latest/instanceGroups/insert. Go with whatever sounds right to you.

DawidJanczak:
I think I'll leave it as is, so feel free to merge. I'm unsure about 2, but at least it seems like the behaviour is as it was before.

PR #254, opened 2017-11-06
Upgrade Cloud SQL to new google api client

emilymye:
Sorry for large CL in advance. I still need to add support for Databases, which weren't a thing in v3, but I think it can wait until after we finish compute since it wasn't a feature in v3.


emilymye:
SG!

PR #255, opened 2017-11-07
Upgrade Compute projects and regions

emilymye:


emilymye:
Added a fix commit - should be good now

PR #256, opened 2017-11-07
Remove deprecated resource/region/zone views

emilymye:
We will need to add managed instance groups, which I think replaced resource views, but we can probably hold off until later.


emilymye:
👍

PR #262, opened 2017-11-10
Upgrade Compute zones

emilymye:
Mostly an update for Zone model.


emilymye:
Whoops, forgot to reply, lgtm

PR #269, opened 2017-11-28
Upgrade Compute firewalls

emilymye:


lib/fog/compute/google/requests/insert_firewall.rb
```diff
@@ -2,40 +2,61 @@ module Fog
...
+        # @option opts [Array<String>] source_ranges
+        # @option opts [Array<String>] source_service_accounts
+        # @option opts [Array<String>] source_tags
+        # @option opts [Array<String>] target_service_accounts
+        # @option opts [Array<String>] target_tags
+        #
+        # @see https://cloud.google.com/compute/docs/reference/latest/firewalls/insert
+        def insert_firewall(firewall_name, opts = {})
+          if opts.key? :network
+            unless opts[:network].start_with?("http://", "https://", "projects/", "global/")
```
icco: should we also check to make sure it's not empty string?

lib/fog/compute/google/requests/insert_firewall.rb
```diff
@@ -2,40 +2,61 @@ module Fog
...
+              opts[:network] = "projects/#{@project}/global/networks/#{opts[:network]}"
+            end
           end
 
-          request(api_method, parameters, body_object)
+          @compute.insert_firewall(
+            @project,
+            ::Google::Apis::ComputeV1::Firewall.new(
+              opts.select { |k, _| INSERTABLE_FIREWALL_FIELDS.include? k }
```
icco: could you do this opts filtering on a seperate line, outside of the function call? I think it'd be cleaner.

lib/fog/compute/google/requests/patch_firewall.rb
```diff
@@ -0,0 +1,26 @@
...
+      class Real
+        ##
+        # Patch a Firewall resource. Supports PATCH semantics.
+        #
+        # @see https://cloud.google.com/compute/docs/reference/latest/firewalls/patch
+        def patch_firewall(firewall_name, options = {})
+          @compute.patch_firewall(
+            @project, firewall_name,
+            ::Google::Apis::ComputeV1::Firewall.new(
+              options.select { |k, _| UPDATABLE_FIREWALL_FIELDS.include? k }
```
icco: It's weird that this is referenced here, despite being defined in a different request.

lib/fog/compute/google/requests/insert_firewall.rb
emilymye: Done.

lib/fog/compute/google/requests/patch_firewall.rb
emilymye: I don't actually know where I should put shared constants for a set of requests . Is compute/google.rb a good place?  I wasn't sure since this just applies to firewalls.

lib/fog/compute/google/requests/insert_firewall.rb
emilymye: Done!

PR #270, opened 2017-11-30
Upgrade forwarding rules (regional and global)

emilymye:


lib/fog/compute/google/models/forwarding_rule.rb
```diff
@@ -4,58 +4,56 @@ class Google
...
         attribute :ip_address, :aliases => "IPAddress"
         attribute :ip_protocol, :aliases => "IPProtocol"
+        attribute :backend_service, :aliases => "backendService"
+        attribute :creation_timestamp, :aliases => "creationTimestamp"
+        attribute :description, :aliases => "description"
+        attribute :id, :aliases => "id"
+        attribute :ip_version, :aliases => "ipVersion"
+        attribute :kind, :aliases => "kind"
+        attribute :load_balancing_scheme, :aliases => "loadBalancingScheme"
+        attribute :network, :aliases => "network"
```
icco: we don't need aliases for stuff that has no case variations

emilymye: Removed!

PR #271, opened 2017-11-30
Upgrade machine type requests

emilymye:
Should I also add models?


icco:
Yeah, that'd be great!

emilymye:
👍

PR #272, opened 2017-11-30
Upgrade existing Compute networks models & requests

emilymye:
I'll probably hold off on adding new requests until we have everything else updated.

### DIFF
--- a/lib/fog/compute/google/models/network.rb
+++ b/lib/fog/compute/google/models/network.rb
@@ -11,8 +11,8 @@ module Fog
         attribute :auto_create_subnetworks, aliases => "autoCreateSubnetworks"
         attribute :creation_timestamp, aliases => "creationTimestamp"
         attribute :description
-        attribute :gateway_i_pv4, aliases => "gatewayIPv4"
-        attribute :i_pv4_range, aliases => "IPv4Range"
+        attribute :gateway_ipv4, aliases => %w(gateway_i_pv4 gatewayIPv4)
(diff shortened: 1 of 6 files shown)


lib/fog/compute/google/models/network.rb
```diff
@@ -8,19 +8,24 @@ class Google
...
-        attribute :kind
-        attribute :id
-        attribute :ipv4_range, :aliases => "IPv4Range"
-        attribute :creation_timestamp, :aliases => "creationTimestamp"
+        attribute :auto_create_subnetworks, aliases => "autoCreateSubnetworks"
+        attribute :creation_timestamp, aliases => "creationTimestamp"
         attribute :description
-        attribute :gateway_ipv4, :aliases => "gatewayIPv4"
-        attribute :self_link, :aliases => "selfLink"
+        attribute :gateway_i_pv4, aliases => "gatewayIPv4"
```
icco: Anyway we can keep `ip` together? Or is this Rubocop complaining?

emilymye: I changed it to match the argument in the Google Client API. I'm not sure if we want it to vary - i could add it as an alias? Mostly we would have to do some conversion during insert.

icco: Ah ok. Let us do an alias. If it's an alias, I think the conversion should happen automatically.

emilymye: done!

PR #274, opened 2017-12-04
Upgrade global addresses and add list request

emilymye:
@icco It doesn't look like I have permissions to request a review (no button in the sidebar)


icco:
Weird, that's annoying you can't request. Whelp I'll keep an eye out. Thanks for everything!

PR #275, opened 2017-12-06
Upgrade compute images and add deprecate_images request. 

emilymye:


lib/fog/compute/google/models/image.rb
```diff
@@ -4,40 +4,49 @@ class Google
...
 
-        # A RawDisk, e.g. -
-        # {
-        #   :source         => url_to_gcs_file,
-        #   :container_type => 'TAR',
-        #   :sha1Checksum   => ,
-        # }
-        attribute :raw_disk, :aliases => "rawDisk"
-
         def preferred_kernel=(_args)
```
icco: We can probably just delete the perferred_kernel calls since they have been deprecated for a while.

lib/fog/compute/google/models/images.rb
```diff
@@ -8,73 +8,72 @@ class Images < Fog::Collection
...
-          "coreos-cloud",
-          "debian-cloud",
-          "google-containers",
-          "opensuse-cloud",
-          "rhel-cloud",
-          "suse-cloud",
-          "ubuntu-os-cloud",
-          "windows-cloud"
-        ].freeze
+        GLOBAL_PROJECTS = %w(
```
icco: Awesome. Thanks for updating this. One thing I've wanted for a long time is an API call that provides these projects.

emilymye: Yeah, that would be nice. I can look into if we can get a method for it or if it's just poorly documented.

lib/fog/compute/google/models/image.rb
emilymye: Done!

PR #288, opened 2018-01-08
method :flatMap not resolved

webervin:
as method is not resolved, no server has external IP returned. 
Probably intended method was:  http://ruby-doc.org/core-2.0.0/Enumerable.html#method-i-flat_map
 
as workaround, until this is merged and released users can use alias_method:

```
module Enumerable
  alias_method :flatMap, :flat_map
end
```


icco:
Thanks!

Issue #289, opened 2018-01-08
Invalid upload source when using Paperclip


sonicdoe:
Using fog-google v1.0.0, I receive a `Google::Apis::ClientError (Invalid upload source)` error when uploading a file using [Paperclip](https://github.com/thoughtbot/paperclip). This seems to be a regression of https://github.com/fog/fog-google/pull/237.

In [`put_object.rb`](https://github.com/fog/fog-google/blob/2d4f0b2cd6d0f901121aafbc7146f4370cc7734c/lib/fog/storage/google_json/requests/put_object.rb#L49-L59), `data.path` will only be used if `data` responds to both `:content_type` and `:path` _and_ if `options[:content_type]` is falsy:

```rb
unless options[:content_type]
  if data.is_a?(String)
    data = StringIO.new(data)
    options[:content_type] = "text/plain"
  elsif data.is_a?(::File)
    options[:content_type] = Fog::Storage.parse_data(data)[:headers]["Content-Type"]
  elsif data.respond_to?(:content_type) && data.respond_to?(:path)
    options[:content_type] = data.content_type
    data = data.path
  end
end
```

Although this behavior hasn’t really changed in v1.0.0, `unless options["Content-Type"]` was changed to `unless options[:content_type]` in [`da9619b`](https://github.com/fog/fog-google/commit/da9619b34274e8112674e1ddced2392f1c5aa4f7#diff-639c71daef71f785b029dcb28455bbe8L21). Therefore, as far as I can tell, https://github.com/fog/fog-google/pull/237 only unintentionally worked because the option was actually passed as `contentType` instead of `Content-Type` (see [`da9619b`](https://github.com/fog/fog-google/commit/da9619b34274e8112674e1ddced2392f1c5aa4f7#diff-02a888ff0b756f121e1280a7935cb5d2L62)).

I’d suggest always using the `data`’s `path` and `content_type` if it responds to these methods by moving the if block out of the `unless options[:content_type]` block.

Note: I have only reproduced this in one Rails app at work and haven’t fully tested it in a fresh Rails app so far (which is also why I haven’t opened a pull request yet).

icco:
@sonicdoe what you're saying seems reasonable. If you can find any other example of this failing (in a simple rails app or somewhere else), I think that'd be enough for me.

icco:
Thanks for reporting and fixing @sonicdoe !

PR #290, opened 2018-01-10
Fix file uploading when using Paperclip

sonicdoe:
[Paperclip](https://github.com/thoughtbot/paperclip) passes `:content_type` to fog which caused fog-google not to use the data’s path (and therefore causing a `Google::Apis::ClientError`). See #289 for more information.

### DIFF
--- a/lib/fog/storage/google_json/requests/put_object.rb
+++ b/lib/fog/storage/google_json/requests/put_object.rb
@@ -52,10 +52,13 @@ module Fog
               options[:content_type] = "text/plain"
             elsif data.is_a?(::File)
               options[:content_type] = Fog::Storage.parse_data(data)[:headers]["Content-Type"]
-            elsif data.respond_to?(:content_type) && data.respond_to?(:path)
-              options[:content_type] = data.content_type
-              data = data.path
             end
+          end
+
+          # Paperclip::AbstractAdapter
+          if data.respond_to?(:content_type) && data.respond_to?(:path)
+            options[:content_type] = data.content_type
+            data = data.path
           end
 
           object_config = ::Google::Apis::StorageV1::Object.new(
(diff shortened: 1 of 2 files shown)


lib/fog/storage/google_json/requests/put_object.rb
```diff
@@ -52,12 +52,14 @@ def put_object(bucket_name,
...
             elsif data.is_a?(::File)
               options[:content_type] = Fog::Storage.parse_data(data)[:headers]["Content-Type"]
-            elsif data.respond_to?(:content_type) && data.respond_to?(:path)
-              options[:content_type] = data.content_type
-              data = data.path
             end
           end
 
+          if data.respond_to?(:content_type) && data.respond_to?(:path)
+            options[:content_type] = data.content_type
```
icco: Random thought, should we not overwrite content_type in this case? 

Also, mind putting a comment right above this `if` saying that this is for Paperclip? Just so it doesn't get erased.

sonicdoe: > should we not overwrite content_type in this case?

I’m not sure I’m following. Do you mean we should prefer `options[:content_type]` over `data.content_type`?

> mind putting a comment right above this `if` saying that this is for Paperclip?

Done 👍

icco: @sonicdoe Yeah, I mean, the user in this case is passing in a content type, which kind of implies to me they want that to be the content type, despite whatever the file actually thinks it is. We could do something like `options[:content_type] ||= data.content_type` to make sure there is always a content type, but default to the one the user passes in.

sonicdoe: You’re right, this does indeed make more sense. It likely won’t matter in practice since this part is probably only used by Paperclip (which always passes the file’s content type).

Anyway, I’ve changed it to `||=` in d814259e2e02ecb93522e755f9efa8c5b7224748 👍

icco:
Awesome! Thanks so much!

PR #292, opened 2018-01-19
Add block back into storage get_object

### LINKED ISSUE #291: `files.get` interface differs significantly between Fog::Storage::GoogleJson and Fog::Storage::GoogleXML
We were trying to consume the new JSON API functionality provided in the latest version (`1.0.1`) of this gem. 

Calls to `files.get` that use the new `GoogleJson` version of `get_object` are not working with our existing code that passes the `files.get` method a block that we use to write the file out to disk. You can see an example of this [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/d9a5c6a2152645749de6996b7fef76dd38b2be90/lib/cloud_controller/blobstore/fog/fog_client.rb#L45-L48).

In the old `GoogleXML` implementation of this, there is apparent [handling of the passed block](https://github.com/fog/fog-google/blob/031a9a941ebde6fd4bf6ba11d3d8a5711a6233c9/lib/fog/storage/google_xml/requests/get_object.rb#L43), whereas in the the new `GoogleJson` implementation [it appears to just return the object directly](https://github.com/fog/fog-google/blob/031a9a941ebde6fd4bf6ba11d3d8a5711a6233c9/lib/fog/storage/google_json/requests/get_object.rb#L65) regardless of whether a block was provided in the call to [files.get](https://github.com/fog/fog-google/blob/031a9a941ebde6fd4bf6ba11d3d8a5711a6233c9/lib/fog/storage/google_json/models/files.rb#L36).

Comments in the new code imply that we need to call this method differently and supply a `download_dest` option.

Unfortunately, this seems like an inconsistency between the two implementations and makes it more difficult to interoperate with `fog-aws` which [functions similarly](https://github.com/fog/fog-aws/blob/d828888b7a4dc3374382090afa16839442aa6d53/lib/fog/aws/requests/storage/get_object.rb#L53) to the `GoogleXML` version. It also makes it very difficult to interoperate with both the `GoogleXML` and `GoogleJson` versions themselves since it's not clear to consumers which codepath the gem will end up using. In our application we allow users to provide arbitrary fog connection config so we want to be able to rely on a standardized interface.

Thanks!
Tim Downey && @ericpromislow

emilymye:
Fixes #291 

tbh the 'streaming' part could be improved but this will work for now

### DIFF
--- a/lib/fog/storage/google_json/requests/get_object.rb
+++ b/lib/fog/storage/google_json/requests/get_object.rb
@@ -23,6 +23,9 @@ module Fog
         #   Set of properties to return
         # @param options [Hash]
         #   Request-specific options
+        # @param &block [Proc]
+        #   Block to pass a streamed object response to. Expected format is
+        #   same as Excon :response_block ({ |chunk, remaining_bytes, total_bytes| ... })
         # @return [Hash] Object metadata with :body attribute set to contents of object
         def get_object(bucket_name, object_name,
                        generation: nil,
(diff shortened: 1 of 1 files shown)


icco:
Wonderful, thank you! Could you change the model to pass the block in?

emilymye:
Are you talking about another model than files.rb? It was actually already being passed in: 

https://github.com/fog/fog-google/blob/031a9a941ebde6fd4bf6ba11d3d8a5711a6233c9/lib/fog/storage/google_json/models/files.rb#L34-L40

icco:
@emilymye whoops, totally was misreading. Thanks!

PR #293, opened 2018-01-22
Fix GoogleJson directories.get signature

icco:
Updates the json directories.get interface signature to match xml

for #291


icco:
cc @tcdowney + @emilymye

PR #299, opened 2018-01-29
Fix arguments for storage list_objects and enable :prefix

### LINKED ISSUE #296: GoogleJSON directories#get does not accept a prefix option
In the older [GoogleXML implementation](https://github.com/fog/fog-google/blob/2d8da689c595cca6bf3bbca6f9382e6c2f4cc0e8/lib/fog/storage/google_xml/models/directories.rb#L21) you could pass this call a `prefix` option that would only fetch objects in the bucket that began with that prefix.

For example, this would only return objects directly under the `06824f8f-d185-409f-be75-55dba0a9a904` directory:
```
prefix = ' buildpack_cache/06/82/06824f8f-d185-409f-be75-55dba0a9a904'
connection.directories.get(dir.key, prefix: prefix).files
```

The newer API implementation no longer accepts `prefix`.
https://github.com/fog/fog-google/blob/2d8da689c595cca6bf3bbca6f9382e6c2f4cc0e8/lib/fog/storage/google_json/models/directories.rb#L12

Is there a way that support for the `prefix` option could be maintained?

Thanks!
Tim Downey && @maryamklabib

emilymye:
Fixes #296 

@icco The issue above was re-opened since it wasn't actually resolved. Could we possibly get a new release for this asap after @tcdowney confirms?

### DIFF
--- a/lib/fog/storage/google_json/models/directories.rb
+++ b/lib/fog/storage/google_json/models/directories.rb
@@ -22,7 +22,9 @@ module Fog
           ).to_h
 
           directory = new(data)
-
+          # Because fog-aws accepts these arguments on files at the
+          # directories.get level, we need to preload the directory files
+          # with these attributes here.
           files_attr_names = %i(delimiter page_token max_results prefix)
 
           file_opts = options.select { |k, _| files_attr_names.include? k }
--- a/lib/fog/storage/google_json/models/directories.rb
+++ b/lib/fog/storage/google_json/models/directories.rb
@@ -21,7 +21,13 @@ module Fog
             :projection => projection
           ).to_h
 
-          new(data)
+          directory = new(data)
+
+          files_attr_names = %i(delimiter page_token max_results prefix)
+
+          file_opts = options.select { |k, _| files_attr_names.include? k }
+          directory.files(file_opts)
+          directory
         rescue ::Google::Apis::ClientError => e
           # metageneration check failures returns HTTP 412 Precondition Failed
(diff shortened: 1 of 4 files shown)


icco:
So the problem here is it won't be compatible with their fog-aws call :/

emilymye:
So is the fix to make our call work with directories.get(:prefix?)

icco:
I think so. If it's not possible, this fix makes a lot of sense to me, but if we can, we would be more consistent across APIs. Another valid option is make the other files method calls match ours. Open to your thoughts on the topic.

emilymye:
I did it so you can do either, but the directories call initializes the files with the given attributes. I think this is what AWS does as well.

lib/fog/storage/google_json/models/directories.rb
```diff
@@ -21,7 +21,13 @@ def get(bucket_name, options = {})
             :projection => projection
           ).to_h
 
-          new(data)
+          directory = new(data)
+
+          files_attr_names = %i(delimiter page_token max_results prefix)
+
+          file_opts = options.select { |k, _| files_attr_names.include? k }
+          directory.files(file_opts)
```
icco: Could you leave a comment here as to why we're doing this?

erjohnso:
Thank you Nat!!

icco:
Thanks to Emily. That was a clever solution to this issue.

maryamklabib:
Thanks @emilymye for this solution and to @icco for being on top of it.

Issue #300, opened 2018-02-05
NoMethodError: undefined method `get_object' for Fog::Compute::Google::Disk


seanmalloy:
I receive the below Ruby stack trace when trying to execute the `get_as_boot_disk` method on an object of type Fog::Compute::Google::Disk. I am using fog-google version 1.2.0.

```
NoMethodError: undefined method `get_object' for #<Fog::Compute::Google::Disk:0x00005636e28b6f90>
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/gems/fog-google-1.2.0/lib/fog/compute/google/models/disk.rb:70:in `get_as_boot_disk'
  /home/sean/projects/vagrant-google/lib/vagrant-google/action/run_instance.rb:153:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/warden.rb:34:in `call'
  /home/sean/projects/vagrant-google/lib/vagrant-google/action/warn_ssh_keys.rb:28:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/warden.rb:34:in `call'
  /home/sean/projects/vagrant-google/lib/vagrant-google/action/warn_networks.rb:28:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/warden.rb:34:in `call'
  /home/sean/projects/vagrant-google/lib/vagrant-google/action/sync_folders.rb:33:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/warden.rb:34:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/builtin/provision.rb:80:in `call'
  /home/sean/projects/vagrant-google/vendor/bundle/ruby/2.4.0/bundler/gems/vagrant-8ee7b01855e9/lib/vagrant/action/warden.rb:34:in `call'
[...]

PR #301, opened 2018-02-05
Remove call to nonexistent function

icco:
At some point we deleted this function: https://github.com/fog/fog-google/blame/v0.5.0/lib/fog/compute/google/models/disk.rb#L59-L70, which caused https://github.com/fog/fog-google/issues/300. This is a fix, altho I'm not entirely sure if we should even have `get_as_boot_disk` anymore.


icco:
@emilymye thoughts?

emilymye:
This LGTM to me - I don't really think we should get rid of the method since it seems like someone tried to use it/it might be useful.

PR #304, opened 2018-02-13
Fix insert_server.rb scheduling option

seanmalloy:


seanmalloy:
It looks lijke `::Google::Apis::ComputeV1::ServiceAccount` is being called instead of `Google::Apis::ComputeV1::Scheduling` when the `:scheduling` option is being used.

I have not tested this yet, but it looks like it should be a simple change. I'll try to test this tomorrow.

seanmalloy:
I was able to test this code and it is working.

icco:
Makes sense, thanks!

PR #305, opened 2018-02-19
Fix setting service accounts and scopes on GCE instances

seanmalloy:
Setting service accounts should use an array of hashes instead
of a hash. See the resource.serviceAccounts[] field in the below
Google API document for details.

https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert

### DIFF
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -426,7 +426,7 @@ module Fog
           options = attributes.reject { |_, v| v.nil? }
 
           if service_accounts && service_accounts[0]
-            service_accounts[0].merge!(:scopes => map_scopes(service_accounts[0][:scopes]))
+            service_accounts[0][:scopes] = map_scopes(service_accounts[0][:scopes])
             options[:service_accounts] = service_accounts
           end
 
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -425,10 +425,9 @@ module Fog
 
           options = attributes.reject { |_, v| v.nil? }
 
-          if service_accounts && service_accounts[:scopes]
-            options[:service_accounts] = service_accounts.merge(
-              :scopes => map_scopes(service_accounts[:scopes])
-            )
+          if service_accounts && service_accounts[0]
+            service_accounts[0].merge!(:scopes => map_scopes(service_accounts[0][:scopes]))
+            options[:service_accounts] = service_accounts
           end
 
           if attributes[:external_ip]


lib/fog/compute/google/models/server.rb
```diff
@@ -425,10 +425,9 @@ def save(username: nil, public_key: nil)
 
           options = attributes.reject { |_, v| v.nil? }
 
-          if service_accounts && service_accounts[:scopes]
-            options[:service_accounts] = service_accounts.merge(
-              :scopes => map_scopes(service_accounts[:scopes])
-            )
+          if service_accounts && service_accounts[0]
+            service_accounts[0].merge!(:scopes => map_scopes(service_accounts[0][:scopes]))
```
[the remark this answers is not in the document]
seanmalloy: Fixed

lib/fog/compute/google/models/server.rb
```diff
@@ -425,10 +425,9 @@ def save(username: nil, public_key: nil)
 
           options = attributes.reject { |_, v| v.nil? }
 
-          if service_accounts && service_accounts[:scopes]
-            options[:service_accounts] = service_accounts.merge(
-              :scopes => map_scopes(service_accounts[:scopes])
-            )
+          if service_accounts && service_accounts[0]
+            service_accounts[0][:scopes] = map_scopes(service_accounts[0][:scopes])
+            options[:service_accounts] = service_accounts
```
icco: Since people are possibly already passing in a scopes hash, could we still support that? Maybe modify the request to make sure only one service account is passed in and its in the right format?

seanmalloy: I will test this tonight and see if it works. But from what I recall it seemed like setting the service accounts and scopes was completely broken on the master branch.

I'll dig into it and let you know.

seanmalloy: I will not get to this until this weekend. Sorry for the delay.

seanmalloy: I tried out the latest code from master to see how it behaves. My initial test seems to indicate that simply setting scopes like this does not actually set the service account scopes. When I go into the Google Cloud Console no service account scopes are set on the GCE instance.

```
 {:scopes => ["storage/full"] }
```

Some screen shots from my `pry` session trying to see how this currently works.

screenshot from 2018-02-26 00-02-25

screenshot from 2018-02-26 00-03-51

icco: I think actually I misread this, what I was asking doesn't make sense. I was thinking you were throwing away an option a user was passing in, which is not what is happening here. This actually looks great and I'm going to merge.

PR #306, opened 2018-02-21
Expose Storage pre-signed object delete url

nolith:
This PR exposes pre-signed URLs for Google Cloud Storage

### DIFF
--- a/lib/fog/storage/google_json.rb
+++ b/lib/fog/storage/google_json.rb
@@ -44,6 +44,7 @@ module Fog
       request :copy_object
       request :delete_bucket
       request :delete_object
+      request :delete_object_url
       request :get_bucket
       request :get_bucket_acl
       request :get_object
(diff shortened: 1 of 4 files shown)


icco:
Hi @nolith! Thanks for the PR. This actually looks like it might be better served by modifying https://github.com/fog/fog-google/blob/master/lib/fog/storage/google_json/requests/get_object_https_url.rb to have an optional argument of `method` that defaults to `"GET"`, that way we don't need to create a new file for every single request type. What do you think?

nolith:
Hi @icco thank you.

Looking at the code base I had the impression that each kind of request is supposed to be on a different class loaded with `request`.

This is exactly the same way `get_object`/`get_object_url` and `put_object`/`put_object_url` is implemented.

Adding an optional parameter to `get_object_https_url` it's a bit strange to me because I had the impression that `get_` doesn't mean _getter_ method but it's the name of the remote operation API, which is _GetObject_

nolith:
The [same PR](https://github.com/fog/fog-aws/pull/431) has merged and released by fog-aws.

I would love to remove this monkey-patching from my production code.
Please @icco  let me know what can I do to bring this forward. 🙏 

Thanks

nolith:
No problem. Thank you

PR #307, opened 2018-03-04
Fix compute disk auto delete hash key

seanmalloy:
The correct name is :auto_delete and not :autoDelete.


seanmalloy:
For reference here is the code in the google-api-ruby-client that ultimately uses the `:auto_delete` hash key.

https://github.com/google/google-api-ruby-client/blob/master/generated/google/apis/compute_v1/classes.rb#L1079

icco:
Thanks!

Issue #315, opened 2018-04-02
Instance create workflow is broken


Temikus:
Basic examples no longer work:

```
[2] pry(main)> disk = connection.disks.create(
[2] pry(main)*   :name => "fog-smoke-test-#{Time.now.to_i}",
[2] pry(main)*   :size_gb => 10,
[2] pry(main)*   :zone_name => "us-central1-f",
[2] pry(main)*   :source_image => "debian-8-jessie-v20161215"
[2] pry(main)* )
Google::Apis::ClientError: invalid: Invalid value for field 'sourceImage': 'debian-8-jessie-v20161215'. The URL is malformed.
from /Users/temikus/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/google-api-client-0.19.8/lib/google/apis/core/http_command.rb:218:in `check_status'
```
- This doesn't work due to new google API client no longer accepting source_image in non-url encoded string.

```
[4] pry(main)> server = connection.servers.create(
[4] pry(main)*   :name => "fog-smoke-test-#{Time.now.to_i}",
[4] pry(main)*   :disks => [disk],
[4] pry(main)*   :machine_type => "n1-standard-1",
[4] pry(main)*   :private_key_path => File.expand_path("~/.ssh/id_rsa"),
[4] pry(main)*   :public_key_path => File.expand_path("~/.ssh/id_rsa.pub"),
[4] pry(main)*   :zone_name => "us-central1-f",
[4] pry(main)*   :username => ENV["USER"],
[4] pry(main)*   :tags => ["fog"],
[4] pry(main)*   :service_accounts => %w(sql-admin bigquery https://www.googleapis.com/auth/compute)
[4] pry(main)* )
ArgumentError: zone is required for this operation
from /Users/temikus/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/fog-core-2.1.0/lib/fog/core/attributes.rb:155:in `requires'
```
- This I'm still figuring out.

Temikus:
Ok, so for instance - zone_name alias was removed in https://github.com/fog/fog-google/commit/f14b3b996fa8b5806b142e3260434307e3e8566f

@emilymye - can you let me know why :zone_name alias was removed? Was it because of the method with the same name? I'm asking since I want to make sure I'm not going to break anything if I add it back.

So far I can either:
- Add it back with a deprecation error and deprecate it and all other :zone_name aliases in 3.0. (we have one in disks and instance groups as well)
or
[...]

PR #316, opened 2018-04-02
Fix instance workflow

Temikus:
Fixing the disk part of #315 

Mainly:
- Added logic to get the image from name if the format is not
  compatible with new Google API Client and added a
  user-friendly error if image is specified incorrectly

- Disk autogenerated object already accepts :source_image as an
  option so removing it from request parameters

- Zone is already set in parameters so should not be duped in
  options


Temikus:
Hound is giving me grief and I do not know why. Need to investigate a bit.

PR #317, opened 2018-04-02
Fix instance workflow - take 2

Temikus:
Fixing the disk part of #315, recreating the PR without my changes to rubocop to make sure I'm not causing HoundCI to act up.

Mainly:

Added logic to get the image from name if the format is not
compatible with new Google API Client and added a
user-friendly error if image is specified incorrectly

Disk autogenerated object already accepts :source_image as an
option so removing it from request parameters

Zone is already set in parameters so should not be duped in
options

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: ruby
 cache: bundler
+before_install:
+  # Added gem update due to travis-ci/issues/8969
+  - gem update --system
+  - gem --version
 sudo: false
 dist: trusty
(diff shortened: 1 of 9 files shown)


Temikus:
Yeah, something is weird with Hound. @icco - Do you know if it's been happening for a while?

icco:
Nope re:hound. It was fine last week.

icco:
Thanks! I'll put together a release this weekend and also try and fix hound issues

PR #326, opened 2018-04-23
Add support for using predefined ACLs, fixes #324

### LINKED ISSUE #324: Storage#put_object_acl does not accept canned ACLs
Hi Folks,

I admit my use case is rather strange as I tend to use the lower level API methods and requests as opposed to the models, and I realise I can use the models if I need to. I'm intending on setting an ACL long after an object has been uploaded to the bucket, and so I'd like to avoid downloading the object in order to re-upload it.

However, onto my problem - I am trying to use the `put_object_acl` method/request in the following manner `connection.put_object_acl(bucket_name, object_name, 'public-read')`, which is [officially supported in the XML API documentation](https://cloud.google.com/storage/docs/xml-api/put-object-acls) (see the blue note).

At the moment, the code is not set up to look for a predefined ACL as the third argument, and instead expects a Hash from which it can construct the request body. I would like to use predefined ACLs, as is [supported by the AWS equivalent request](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/put_object_acl.rb#L39). I'm looking at forking the repo at the moment and getting stuck in if you don't mind.

Thanks for reading. :)

-- easkay

easkay:
As per issue #324, I decided to make a start on this functionality and would like to know what folks think. I'm not sure which tests need writing and where, especially since things are migrating to minitest. Some guidance would be appreciated.

Also, the changes introduced here are pretty much entirely copied from https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/put_object_acl.rb#L39.

### DIFF
--- a/lib/fog/storage/google_xml/models/directory.rb
+++ b/lib/fog/storage/google_xml/models/directory.rb
@@ -7,9 +7,8 @@ module Fog
         attribute :creation_date, :aliases => "CreationDate"
 
         def acl=(new_acl)
-          valid_acls = ["private", "public-read", "public-read-write", "authenticated-read"]
-          unless valid_acls.include?(new_acl)
-            raise ArgumentError.new("acl must be one of [#{valid_acls.join(', ')}]")
+          unless Utils::VALID_ACLS.include?(new_acl)
+            raise ArgumentError.new("acl must be one of [#{Utils::VALID_ACLS.join(', ')}]")
           end
           @acl = new_acl
         end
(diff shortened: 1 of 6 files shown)


easkay:
@icco No worries, thanks for taking a look. I'll work on your feedback later this evening and update the PR. I notice there's no mock for this request either, do you have any suggestions on making one? Not a problem if you don't :)

icco:
@easkay I was hoping you could update the File and Directory models as well that are using public acls. As for mocks, usually copying the response from google by making the request with curl and then scrubbing data is how it's been done in the past.

easkay:
@icco Ah, sorry I misunderstood and/or misread. I'll get on that in a few minutes :) Apologies for the misunderstanding.

easkay:
@icco I wasn't entirely sure where these should live. Utils seems like a good place on the surface of it, what do you think?

icco:
Thanks so much!

PR #330, opened 2018-04-26
Fixed example with `predefined_acl`

vimutter:
https://github.com/fog/fog-google/blob/master/lib/fog/storage/google_json/requests/put_bucket.rb#L18


icco:
Thanks!

PR #338, opened 2018-05-22
Fixing SQL

### LINKED ISSUE #337: SQL is broken
- SQL tests do not actually delete resources and are _extremely_ flaky due to the underlying API having issues from time to time combined with how instances are created for tests.

- Methods in sql model do not match other Fog models, nor do they match eachother:
```
        def clone(destination_name, async: true, log_filename: nil, log_position: nil)
        def destroy(async = false)
        def export(uri, options: {})
        def import(uri, options = {})
        def reset_ssl_config(async: true)
        def restart(async: true)
        def restore_backup(backup_run_id, async: true)
```
, notice the positional parameter in `import` and named one in `export`, positional in `destroy`, named in `restart` and `create` is missing sync/async logic altogether.

I'm going to rewrite the models and tests, but this will likely require a major version bump. 
@icco Are you ok bumping to 3.0 when I'm done? 

This should stabilize our CI tests so we can reliably verify changes and I'll bump to new google client to support Ruby 2.5 while we're at it.

Temikus:
A lot of SQL refactoring, mostly tests. This will slow tests down a bit but will fix 99% of the flakiness, and add proper lifecycle tests.

## SQL Tests

- Add new test factories for SQLv1/v2 and SQL Users/SslCerts and instantiated basic collection tests
- Major refactor of SQLv1 and SQLv2 tests, getting rid of old shared helpers and tests that were leaving cruft in test projects and were very flaky, incl.:
  - Adding SQL users factory and migrate user tests to it
  - Adding SslCert factory
  - Breaking out tests by subject collection

## Minor CI/Test improvements
  - Make monitoring cleanup less chatty
  - Bump GoogleSQL instance tier to one with SLA and MakeSQL tests serial
  - Set github status if any CI task failed
  - Tweak rubocop rules
  - Add task to lint changed files

## Drive-by improvements:

- Set SQL delete/create operations to synchronous

- Fixes for SQL Users model workflow
  - If Etag is nil then User doesn't exist, so should be inserted,
    not the other way around.
  - Changed async to positional parameter to match other models.

- Minor refactor of tier tests
  - Remove list test as .all depends on list
  - Scope down the test to not instantiate a full client

- Improve disk type test to include more regions
 

Fixes #337


Temikus:
SQL Api is being flaky again, it seems -_-

I'll try to figure this out, probably will need to wrap with `Retriable` like with Stackdriver...

Temikus:
So I get tons of `Google::Apis::ClientError: invalidState: The instance or operation is not in an appropriate state to handle the request.` even though the operations reported as ready. I've synchronised almost everything. All of those errors are in cleanup BUT it still cleans those up properly 0_o. Probably those resources were not designed to be created/destroyed in 10 minutes.

I'll need to sleep on this... Maybe add a simple wait to teardown or something.

PR #346, opened 2018-05-26
Fix get_health when called with an instance name

bpaquet:
Without that, the Hash[results] does not work as expected.
Another option is to not return the same signature when an instance name is specified (data can be returned directly)


Temikus:
Thank you for this \o/

I've added a test so we don't have this breaking again.

PR #347, opened 2018-05-27
Target pool health test

Temikus:
To prevent #346 from re-occurring.


Temikus:
Last failure can be ignored - it's due to my rebase (doesn't include the fix from the PR).

PR #348, opened 2018-05-29
Add templates and instance group managers

bpaquet:
Sorry, did not write tests, but I'm using it in prod.

### DIFF
--- a/lib/fog/compute/google/models/instance_group_manager.rb
+++ b/lib/fog/compute/google/models/instance_group_manager.rb
@@ -1,0 +1,51 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManager < Fog::Model
+        identity :name
+
+        attribute :kind
+        attribute :self_link, :aliases => "selfLink"
+        attribute :description
+        attribute :zone
+        attribute :region
+        attribute :instance_group, :aliases => "instanceGroup"
+        attribute :instance_template, :aliases => "instanceTemplate"
+        attribute :target_pools, :aliases => "targetPools"
+        attribute :base_instance_name, :aliases => "baseInstanceName"
+        attribute :current_actions, :aliases => "currentActions"
+        attribute :target_size, :aliases => "targetSize"
+        attribute :named_ports, :aliases => "namedPorts"
+
+        def save
+          requires :name, :zone, :base_instance_name, :target_size, :instance_template
+
+          data = service.insert_instance_group_manager(name, zone.split("/")[-1], instance_template, base_instance_name,
+          target_size, target_pools, named_ports, description)
+          operation = Fog::Compute::Google::Operations.new(:service => service).get(data.name, zone.split("/")[-1])
+          operation.wait_for { !pending? }
+          reload
+        end
+
+        def destroy(async = true)
+          requires :name, :zone
+          operation = service.delete_instance_group_manager(name, zone.split("/")[-1])
+          operation.wait_for { ready? } unless async
+          operation
+        end
+
+        def set_instance_template(instance_template)
+          service.set_instance_template self, instance_template
+        end
+
+        def recreate_instances(instances)
+          service.recreate_instances self, instances
+        end
+
+        def abandon_instances(instances)
+          service.abandon_instances self, instances
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/compute/google/models/instance_group_managers.rb
+++ b/lib/fog/compute/google/models/instance_group_managers.rb
@@ -1,0 +1,42 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManagers < Fog::Collection
+        model Fog::Compute::Google::InstanceGroupManager
+
+        def all(zone: nil, filter: nil, max_results: nil,
+                order_by: nil, page_token: nil)
+          opts = {
+            :filter => filter,
+            :max_results => max_results,
+            :order_by => order_by,
+            :page_token => page_token
+          }
+          data = []
+          if zone
+            data += service.list_instance_group_managers(zone, opts).items || []
+          else
+            service.list_aggregated_instance_group_managers(opts).items.each_value do |group|
+              data.concat(group.instance_group_managers) if group.instance_group_managers
+            end
+          end
+
+          load(data.map(&:to_h))
+        end
+
+        def get(identity, zone = nil)
+          if zone
+            if instance_group_manager = service.get_instance_group_manager(identity, zone)
+              new(instance_group_manager.to_h)
+            end
+          else
+            all(:filter => "name eq .*#{identity}").first
+          end
+        rescue ::Google::Apis::ClientError => e
+          raise e unless e.status_code == 404
+          nil
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/compute/google/models/instance_template.rb
+++ b/lib/fog/compute/google/models/instance_template.rb
@@ -1,0 +1,47 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceTemplate < Fog::Model
+        identity :name
+
+        attribute :kind
+        attribute :self_link, :aliases => "selfLink"
+        attribute :description
+        # Properties is a hash describing the templates
+        # A minimal example is
+        # :properties => {
+        #   :machine_type => TEST_MACHINE_TYPE,
+        #   :disks => [{
+        #     :boot => true,
+        #     :initialize_params => {
+        #       :source_image => "projects/ubuntu-os-cloud/global/images/ubuntu-1804-bionic-v20180522"}
+        #     }],
+        #     :network_interfaces => [{
+        #       :network => "global/networks/default"
+        #     }]
+        #   }
+        # }
+        # @see https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert
+        attribute :properties
+
+        def save
+          requires :name
+          requires :properties
+
+          data = service.insert_instance_template(name, properties, description)
+          operation = Fog::Compute::Google::Operations.new(:service => service).get(data.name)
+          operation.wait_for { !pending? }
+          reload
+        end
+
+        def destroy(async = true)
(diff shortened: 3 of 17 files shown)


lib/fog/compute/google/models/instance_group_managers.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManagers < Fog::Collection
+        model Fog::Compute::Google::InstanceGroupManager
+
+        def all(filters = {})
+          if filters[:zone]
+            data = Array(service.list_aggregated_instance_group_managers(filters[:zone]))
```
icco: Why do you need to cast this as an `Array`? What does it return normally?

lib/fog/compute/google/models/instance_group_managers.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManagers < Fog::Collection
+        model Fog::Compute::Google::InstanceGroupManager
+
+        def all(filters = {})
```
icco: `data = []` should be the first line in this function to properly declare scope.

lib/fog/compute/google/models/instance_group_managers.rb
```diff
@@ -0,0 +1,22 @@
...
+          if filters[:zone]
+            data = Array(service.list_aggregated_instance_group_managers(filters[:zone]))
+          else
+            data = []
+            service.list_aggregated_instance_group_managers.items.each_value do |group|
```
icco: I believe this could be simplified to 
```ruby
data = service.list_aggregated_instance_group_managers.items.map {|group| group.instance_group_managers }
```

icco:
Also, feel free to ignore those codeclimate errors.

lib/fog/compute/google/models/instance_group_managers.rb
bpaquet: This is not written like that in others files, but yes, we can do that.

lib/fog/compute/google/models/instance_group_managers.rb
bpaquet: Fixed.

lib/fog/compute/google/models/instance_group_managers.rb
bpaquet: Done.

Temikus:
First of all - thank you!

Let me write in a quick test for this and I’ll check if this fits our standard model/collection lifecycle.

I aim to respond by Wednesday (SYD) at the latest.

lib/fog/compute/google/models/instance_group_managers.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManagers < Fog::Collection
+        model Fog::Compute::Google::InstanceGroupManager
```
Temikus: This is missing a `get` method.

lib/fog/compute/google/models/instance_group_manager.rb
```diff
@@ -0,0 +1,28 @@
+module Fog
+  module Compute
+    class Google
+      class InstanceGroupManager < Fog::Model
```
Temikus: This is missing a `.save` method.

lib/fog/compute/google/requests/insert_instance_template.rb
```diff
@@ -0,0 +1,23 @@
...
+          Fog::Mock.not_implemented
+        end
+      end
+
+      class Real
+        def insert_instance_template(name, properties, description)
+          instance_template = ::Google::Apis::ComputeV1::InstanceTemplate.new(
+            :description => description,
+            :name => name,
+            :properties => properties,
```
Temikus: If I'm reading this correctly, `properties` will need to be an instance of google api client's `InstanceProperties` class.

If so - fog avoids exposing underlying API structures, so parameters need to be provided in hashes, see: https://github.com/fog/fog-google/blob/985cd021bcfe4f403deb7c0077997529d529b4b1/lib/fog/compute/google/requests/insert_server.rb

lib/fog/compute/google/requests/insert_instance_template.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Compute
+    class Google
+      class Mock
+        def insert_instance_template(_name, _properties, _description)
+          Fog::Mock.not_implemented
+        end
+      end
+
+      class Real
+        def insert_instance_template(name, properties, description)
```
Temikus: Since it's not entirely clear what `properties` should contain and in what structure, can you please add a doc comment?

E.g.: https://github.com/fog/fog-google/blob/985cd021bcfe4f403deb7c0077997529d529b4b1/lib/fog/compute/google/requests/insert_server.rb#L38-L60

lib/fog/compute/google/models/instance_templates.rb
```diff
@@ -0,0 +1,22 @@
...
+        def all
+          data = service.list_instance_templates.items
+          load(data.map(&:to_h))
+        end
+
+        def get(identity)
+          if instance_template = service.get_instance_template(identity)
+            new(instance_template.to_h)
+          end
+        rescue Fog::Errors::NotFound
```
Temikus: You should rescue `Google::Apis::ClientError: notFound`, this currently doesn't return nil as it should.

For an example see:
https://github.com/fog/fog-google/blob/985cd021bcfe4f403deb7c0077997529d529b4b1/lib/fog/compute/google/models/servers.rb#L29-L42

Temikus:
@bpaquet to make things easier I've published the quick sanity tests I wrote:
https://github.com/fog/fog-google/pull/350

lib/fog/compute/google/models/instance_templates.rb
bpaquet: Done

lib/fog/compute/google/models/instance_group_managers.rb
bpaquet: Done

lib/fog/compute/google/models/instance_group_manager.rb
bpaquet: Done. Also added a bunch of attributes; reload and destroy.

lib/fog/compute/google/models/instance_template.rb
```diff
@@ -0,0 +1,33 @@
...
+
+          data = service.insert_instance_template(name, properties, description)
+          operation = Fog::Compute::Google::Operations.new(:service => service).get(data.name)
+          operation.wait_for { !pending? }
+          reload
+        end
+
+        def destroy(async = true)
+          requires :name
+          operation = service.delete_instance_template(name)
```
Temikus: You need to return a Fog::Compute::Google::Operation model here, see your `save` method.
Otherwise you'll get a Google API client class that doesn't respond to `.wait_for`:
```
NoMethodError: undefined method `wait_for' for #<Google::Apis::ComputeV1::Operation:0x00007fd00aed3f00>
```

lib/fog/compute/google/requests/insert_instance_group_manager.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+      end
+
+      class Real
+        def insert_instance_group_manager(name, zone, instance_template, base_instance_name,
+                                          target_size, target_pools, named_ports, description)
+          instance_group_manager = ::Google::Apis::ComputeV1::InstanceGroupManager.new(
+            :description => description,
+            :name => name,
+            :instance_template => instance_template.self_link,
```
Temikus: Nit: This assumes that an object is going to be passed in.
Usually models should accept self_links and resource names as well.

lib/fog/compute/google/models/instance_template.rb
```diff
@@ -0,0 +1,33 @@
...
+  module Compute
+    class Google
+      class InstanceTemplate < Fog::Model
+        identity :name
+
+        attribute :kind
+        attribute :self_link, :aliases => "selfLink"
+        attribute :description
+        # Properties is a hash describing the templates
+        # @see https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert
```
Temikus: Nit: If at all possible, add a minimal hash example. The doc itself is quite complicated and it took me a while to figure the right parameters out (didn't work with IGM api before). For newer people that haven't worked with Google API's at all that may be a higher barrier.

I came up with this as a minimal test case:
```
      :properties => {
          :machine_type => TEST_MACHINE_TYPE,
          :disks => [{
                         :boot => true,
                         :initialize_params =>
                          {:source_image => "projects/ubuntu-os-cloud/global/images/ubuntu-1804-bionic-v20180522"}
                     }],
          :network_interfaces => [{ :network => "global/networks/default" }]
      }
```
, but wanted to check with you since you can probably provide a better real-life example.

bpaquet: Done.

No, I'm using template to clone existing template and change the image.

lib/fog/compute/google/models/instance_template.rb
bpaquet: Done

lib/fog/compute/google/requests/insert_instance_group_manager.rb
bpaquet: Done in the three requests.

lib/fog/compute/google/models/instance_group_managers.rb
```diff
@@ -0,0 +1,31 @@
...
+          else
+            service.list_aggregated_instance_group_managers.items.each_value do |group|
+              data.concat(group.instance_group_managers) if group.instance_group_managers
+            end
+          end
+
+          load(data.map(&:to_h))
+        end
+
+        def get(identity, zone)
```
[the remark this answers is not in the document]
bpaquet: Done.

Issue #352, opened 2018-06-05
Models should not have more than 1 required parameter in get()


Temikus:
This will need to be fixed, otherwise lifecycle doesn't match:
```
λ grep -R "def get([a-z_\*-]*, " . 
./lib/fog/google/models/sql/ssl_certs.rb:        def get(instance_id, sha1_fingerprint)
./lib/fog/google/models/sql/backup_runs.rb:        def get(instance_id, backup_run_id)
./lib/fog/storage/google_xml/models/directories.rb:        def get(key, options = {})
./lib/fog/storage/google_xml/models/files.rb:        def get(key, options = {}, &block)
./lib/fog/storage/google_json/models/directories.rb:        def get(bucket_name, options = {})
./lib/fog/storage/google_json/models/files.rb:        def get(key, options = {}, &block)
./lib/fog/compute/google/models/target_pools.rb:        def get(identity, region = nil)
./lib/fog/compute/google/models/addresses.rb:        def get(identity, region)
./lib/fog/compute/google/models/instance_groups.rb:        def get(identity, zone = nil)
./lib/fog/compute/google/models/images.rb:        def get(identity, project = nil)
./lib/fog/compute/google/models/instance_group_managers.rb:        def get(identity, zone = nil)
./lib/fog/compute/google/models/forwarding_rules.rb:        def get(identity, region = nil)
./lib/fog/compute/google/models/operations.rb:        def get(identity, zone = nil, region = nil)
./lib/fog/compute/google/models/disks.rb:        def get(identity, zone)
./lib/fog/compute/google/models/servers.rb:        def get(identity, zone = nil)
./lib/fog/compute/google/models/machine_types.rb:        def get(identity, zone)
./lib/fog/compute/google/models/disk_types.rb:        def get(identity, zone = nil)
./lib/fog/compute/google/models/target_instances.rb:        def get(target_instance, zone = nil)
./lib/fog/compute/google/models/subnetworks.rb:        def get(identity, region)
./lib/fog/dns/google/models/records.rb:        def get(name, type)
```

Game plan:
- [x] Unit tests for basic model lifecycle methods.
- [x] Add lookup logic if the resource location is not provided (probably via `all`?)
- [ ] Switch all zone/region parameters to be optional, but leave them in to speed up logic when needed.
- [ ] Fixup the regression in collection tests (see #396)
- [x] Unit tests checking for get() required parameters so this doesn't deviate again.
- [ ] Remove `skip` test fixtures in `test:unit`

[...]

PR #353, opened 2018-06-06
Adding unit tests

Temikus:
This should help us keep models/collections the same across the board and avoid issues like #352 and help newcomers iterate quickly to avoid us needing to point out missing methods like in #348


Temikus:
Wohoo \o/ Unit tests are successfully running on Travis.

Now just to fix the `Projects` collection.

Temikus:
This should now be done.

icco:
Thanks!

Issue #365, opened 2018-06-22
Some save() methods have arguments and shouldn't


Temikus:
File is probably fine but user/server should _not_ have them
```
λ grep -R 'def save([a-z ]' .
./lib/fog/google/models/sql/user.rb:        def save(password: nil)
./lib/fog/storage/google_xml/models/file.rb:        def save(options = {})
./lib/fog/compute/google/models/server.rb:        def save(username: nil, public_key: nil)
```

Issue #362, opened 2018-06-21
Unable to add ssh-key when provisioning a new instance


EUGinsane:
Hi all,
When new instance with:
```
 server = connection.servers.create(
    :name => "fog-smoke-test-#{Time.now.to_i}",
    :disks => [disk],
    :machine_type => "n1-standard-1",
    :private_key_path => File.expand_path("~/.ssh/id_rsa"),
    :public_key_path => File.expand_path("~/.ssh/id_rsa.pub"),
    :zone => "us-central1-f",
    :username => ENV["USER"],
    :tags => ["fog"],
    :service_accounts => { :scopes => %w(sql-admin bigquery https://www.googleapis.com/auth/compute) }
  )
```

I generated SSH key and new instance is suscess but I dont see any ssh key for VM instance. Please help me recheck. 

![Uploading ssh_key.PNG…]()
Thanks

Temikus:
@EUGinsane Thank you for filing this!!! I found a regression, will fix it soon. 
Until then you can workaround by doing this after create:

```
server.add_ssh_key(ENV["USER"], File.expand_path("~/.ssh/id_rsa.pub"))
```

Apologies for the hassle m(_ _)m

Temikus:
Fix has been pushed, this will be released with `1.6.0`

EUGinsane:
Thanks @Temikus ,
I want to ask one question. When I created Windows instance, how to connection to this vm ?

Temikus:
@EUGinsane This isn't implemented directly in fog, I'm afraid. 
There are 2 ways to connect - PowerShell and RDP[1]

If you go RDP route you can add an exec() in your script/app that will call the rdp client of choice.
Concerning powershell, I'm not sure.

[1] https://cloud.google.com/compute/docs/instances/connecting-to-instance

PR #366, opened 2018-06-22
Fixing Fog::Server user/sshkey workflow

Temikus:
For some reason those were abandoned in 1.0, so core examples and `bootstrap` no longer worked.
E.g. the following no longer provisioned the keys properly:

```
 server = connection.servers.create(
    :name => "fog-smoke-test-#{Time.now.to_i}",
    :disks => [disk],
    :machine_type => "n1-standard-1",
    :private_key_path => File.expand_path("~/.ssh/id_rsa"),
    :public_key_path => File.expand_path("~/.ssh/id_rsa.pub"),
    :zone => "us-central1-f",
    :username => ENV["USER"],
    :tags => ["fog"],
    :service_accounts => { :scopes => %w(sql-admin bigquery https://www.googleapis.com/auth/compute) }
  )
```
This patch will make sure `:private_key_path` and `:public_key_path` are processed properly again.

Fixing #362 

Not even mentioning that `save()` shouldn't have had arguments in the first place.
But that's going to be fixed by a larger effort in 2.0 #365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Next release (1.6.0)
+
+### User-facing
+
+#### Added
+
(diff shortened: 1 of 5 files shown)


Temikus:
Added a unit test so we don't have to go through this again.

Temikus:
Now instance manager group tests are failing for no reason :|
I'll take a look and see if something's happening to the API.

Temikus:
All tests pass, merging.

PR #367, opened 2018-06-23
Bootstrap autodelete

### LINKED ISSUE #17: servers::bootstrap should use disk/autoDelete
Right now, bootstrapped servers automatically create a disk to use with the server, but do not automatically destroy that disk when the server is destroyed.  I propose we use the disk/autoDelete option when creating bootstrapped instances, to avoid orphaned disks, (more info [here](https://cloud.google.com/compute/docs/instances#startinstanceapi)).

@icco You know this codebase better than I do; objections?  Thanks.

Temikus:
- Add tests for `Fog::Compute::Google::Server.bootstrap`
- Set disks to autodelete for servers created with `Fog::Compute::Google::Server.bootstrap`

+ some drive-by fixes to `.bootstrap`

Fixes #17


Temikus:
All tests pass \o/ 

Merging.

PR #372, opened 2018-06-25
Add global addresses tests

Temikus:
To verify whether the coverage delta is done properly, I'm adding more tests for a previously untested resource. Let's see if there's any delta.

Bonus: more tests \o/


Temikus:
Alright, pushed our bot account here. Let's see if this works.

PR #375, opened 2018-06-26
Add timeout options (Google::Apis::ClientOptions) to GoogleJSON API

dosuken123:
Currently, fog-google doesn't allow users to set timeout options when they upload objects to GCS.

This PR extends the support.

Here, in GitLab's usage, we can pass those params as the followings.

```
# gitlab.yml

connection:
  provider: Google
  google_project: xxx
  google_client_email: xxx
  google_json_key_location: xxx
  open_timeout_sec: 111
  read_timeout_sec: 123
  send_timeout_sec: 321
```

### DIFF
--- a/lib/fog/storage/google_json.rb
+++ b/lib/fog/storage/google_json.rb
@@ -16,7 +16,10 @@ module Fog
         :google_key_location,
         :google_key_string,
         :google_json_key_location,
-        :google_json_key_string
+        :google_json_key_string,
+        :open_timeout_sec,
+        :read_timeout_sec,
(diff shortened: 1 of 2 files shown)


Temikus:
@dosuken123 Thanks for the contribution! 

LGTM, let me just run the test suite over this.

Temikus:
LGTM. All storage tests pass.

Only Stackdriver CI failed due to Stackdriver API being flaky, but that's unrelated.
I'll investigate that separately in #378

@dosuken123 Thank you for the contribution! Do you want me to do a release right away?

PR #393, opened 2018-07-14
More compute coverage and some small improvements

Temikus:
Server is our most complicated model but it has the worst test coverage.
This is trying to improve it and fix the docs while I'm at it.

Due to the need to instantiate a bunch of servers during the test this slows them down quite a bit...
Looking at ballpark figures - 73 minutes :|

I'll start thinking about speeding things up. I think we can parallelise things more, but that requires careful planning of cleanup (since different API's can trump eachother).


Temikus:
This PR is broken - hooks do not work... Reinstantiating.

PR #397, opened 2018-07-16
More parallelization

Temikus:
Our pipeline needs to be fanned out a bit so tests take less time.
This is the first attempt on that.


Temikus:
@icco You were asking before to clean up our CI images situation.

We now have a proper DockerHub org for fog/google and images are pulled from it: https://hub.docker.com/r/rubyfog/fog-google/

You should be one of the owners - if something's not working - just let me know.

Temikus:
Alright. Now our tests run for 40-something minutes instead of 2+ hours with 84%+ coverage \o/

screen shot 2018-07-17 at 13 09 14

Temikus:
Alright, updated prod env. One last run and we can merge.

PR #400, opened 2018-07-19
Fixing empty get() behavior

### LINKED ISSUE #33: For some resources, resource#get(nil) returns a resource
I would expect that, for example, `Fog::Compute[:google].servers.get(nil)` should either return `nil` or raise an error.  Instead, if at least one server exists in my project,

``` ruby
> Fog::Compute[:google].servers.get(nil)
=> <Fog::Compute::Google::Server
      name="server-name",
      ...
    >
```

This is because of the way we find servers, disks, and other resources that are zoned/regioned.  For example:

``` ruby
servers = service.list_aggregated_servers(:filter => "name eq .*#{identity}").body['items']
```

Every server matches that filter if `identity` is `nil`.

Temikus:
Fixes #33

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,9 @@ Style/ConditionalAssignment:
   SingleLineConditionsOnly: true
   EnforcedStyle: assign_inside_condition
 
+Style/RedundantReturn:
+  Enabled: false
+
 Style/FormatString:
   Enabled: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ matrix:
     - rvm: 2.5
     - rvm: jruby-head
   allow_failures:
-  # allow 2.5 to fail until Travis CI is fixed.
-  # See travis-ci/travis-ci#8969
-    - rvm: 2.5
     - rvm: jruby-head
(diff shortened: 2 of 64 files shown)


Temikus:
Holy moly this is complicated.

So far I see 6 kinds of behavior:
- Selecting the first instance from `.all` listing
- Returning nil
- Returning a weird nil-object, e.g.:
```
<Fog::Compute::Google::Network
            name=nil,
            auto_create_subnetworks=nil,
            creation_timestamp=nil,
            description=nil,
            gateway_i_pv4=nil,
            i_pv4_range=nil,
            id=0,
            kind="compute#networkList",
            peerings=nil,
            routing_config=nil,
            self_link="https://www.googleapis.com/compute/v1/projects/REDACTED/global/networks/",
            subnetworks=nil
          >
```
- Google::Apis::ClientError
- Miscellaneous `NoMethodError` bubbling up from the methods
- ArgumentError - wrong number of arguments

I'll try to tackle this one by one but not all of them can be fully fixed without a major version release.

Temikus:
Cleared up the comments from the no longer relevant linter. Otherwise hard to see what else I missed.

PR #408, opened 2018-07-26
Update supported ruby versions

### LINKED ISSUE #407: Clarify supported ruby versions
Today I've discovered that we no longer work on ruby 2.1 while reproducing the issue:
```
λ bundle                                                                                                                                   (5)
Fetching gem metadata from https://rubygems.org/........
RubyGems 2.0.3 is not threadsafe, so your gems will be installed one at a time. Upgrade to RubyGems 2.1.0 or higher to enable parallel gem installation.
Fetching rake 12.3.1
Installing rake 12.3.1
Fetching public_suffix 3.0.2
Installing public_suffix 3.0.2
Gem::InstallError: public_suffix requires Ruby version >= 2.1.
An error occurred while installing public_suffix (3.0.2), and Bundler cannot continue.
Make sure that `gem install public_suffix -v '3.0.2' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  fog-google was resolved to 1.6.0, which depends on
    google-api-client was resolved to 0.23.0, which depends on
      googleauth was resolved to 0.6.2, which depends on
        signet was resolved to 0.8.1, which depends on
          addressable was resolved to 2.5.2, which depends on
            public_suffix
```
Digging into it we have the following set of constraints:
- Our dependency (`google-api-client`) is not supporting deprecated ruby versions.
- Our CI is testing API compatibility starting ruby-2.3

@icco Are you ok if I clarify it in the README that we support Ruby 2.3+ and in general strive to support the current supported versions of Ruby?

I want to make sure we do our users justice so your opinion on this is very welcome.

Temikus:
@icco - PTAL the wording, want to make sure it sounds fair.

Fixes #407

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,7 @@ The main maintainers for the Google sections are @icco, @Temikus and @plribeiro3
 
 ## Important notices
 
- - As of **v0.1.1**, Google no longer supports Ruby versions less than 2.0.0.  
-
- - As of **v1.0.0**, fog-google includes google-api-client as a dependency, there is no need to include it separately anymore.
+- As of **v1.0.0**, fog-google includes google-api-client as a dependency, there is no need to include it separately anymore.
 
 See **[MIGRATING.md](MIGRATING.md)** for migration between major versions.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,6 +112,15 @@ $ bundle exec pry
     kind="compute#instance",
 ```
 
+## Supported Ruby Versions
```
icco: This makes sense, and reads well, but I'd put it as an h3 under Important Notices at the top, as this is pretty far down the page.

Temikus: Ack, added a notice. Don't want to put the whole paragraph up there though, so created a link.

PR #411, opened 2018-08-19
Make HTTPS the default for XML requests

### LINKED ISSUE #303: GoogleXML::File should support private HTTPS url
# Problem
When using `GoogleXML::File.url`, the returned URL is always using the `http://` scheme. This will cause insecure redirects when trying to forward a request to the private URL of an object.

# Solution
The scheme should be configurable across all URL generation methods, so that you can serve HTTP/HTTPS urls from either private or public URLs.

See https://github.com/fog/fog-google/blob/30d6b447399146f1ed04daf35dd8d4bedffbd146/lib/fog/storage/google_xml/models/file.rb#L132

stanhu:
Closes #303

### DIFF
--- a/lib/fog/storage/google_xml/models/file.rb
+++ b/lib/fog/storage/google_xml/models/file.rb
@@ -118,7 +118,7 @@ module Fog
 
         def url(expires, options = {})
           requires :key
-          collection.get_http_url(key, expires, options)
+          collection.get_https_url(key, expires, options)
         end
 
         private


lib/fog/storage/google_xml/models/file.rb
```diff
@@ -118,7 +118,7 @@ def save(options = {})
 
         def url(expires, options = {})
           requires :key
-          collection.get_http_url(key, expires, options)
+          collection.get_https_url(key, expires, options)
```
stanhu: Hmm, we may want to support the `:scheme` option after all.

stanhu: Perhaps not, given `get_object_url` is deprecated in favor of `get_object_https_url`.

Temikus:
@stanhu Thank you for the PR

This is a breaking change so I would avoid changing out the scheme without notice, if at all possible until we release 2.0.

However, I do understand there may be some dependency interface (e.g. Paperclip) that I'm not aware of that uses it - can you give me some context on this?

Issue #417, opened 2018-09-06
The format Fog::Storage::Google is deprecated


HarlemSquirrel:
Upon updating fog-core 2.1.2 (was 2.1.0), I received this warnings while running my test suite:

```
[fog][DEPRECATION] Unable to load Fog::Google::Compute
[fog][DEPRECATION] The format Fog::Compute::Google is deprecated
[fog][DEPRECATION] Unable to load Fog::Google::DNS
[fog][DEPRECATION] The format Fog::DNS::Google is deprecated
[fog][DEPRECATION] Unable to load Fog::Google::Storage
[fog][DEPRECATION] The format Fog::Storage::Google is deprecated
[fog][DEPRECATION] Unable to require fog/google/storage
[fog][DEPRECATION] The format fog/storage/google is deprecated
[fog][DEPRECATION] Unable to load Fog::Google::Storage
[fog][DEPRECATION] The format Fog::Storage::Google is deprecated
```

Adding `gem 'fog-core', '2.1.0'` to my Gemfile fixed this for me.

Temikus:
Thanks for reporting this! I'll shortly release a version locking the dep and start working on sorting this out the proper way.

This is a good time to lock the rest of the fog core deps down as well.

stanhu:
Thanks, the fog-core changes are really disruptive. The latest version of fog-aws uses fog-core 2.1.2+, and the warning messages are alarming if you go with the lowest common denominator and use fog-core 2.1.0 with fog-aws: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/21788#note_102836639

Temikus:
1.8.0 has been released. 

If you're still seeing any issues - just let me know.

stanhu:
@Temikus Thanks, first issue is that the fog-core version dependency wasn't bumped. :) https://github.com/fog/fog-google/pull/427

codeodor:
> Adding gem 'fog-core', '2.1.0' to my Gemfile fixed this for me.

Is there a way to do that without getting all of the subcomponents?

majksner:
I'm getting this warning again, using the latest version of fog-google. I can provide any additional info if necessary.

Temikus:
@majksner this was an intentional tradeoff, see #562

VILATTI:
+1

PR #419, opened 2018-09-17
Fix core dep

Temikus:
This should alleviate #417 until we can properly upgrade all the tests, etc.

Additionally, locks the major versions of other upstream deps.


DawidJanczak:
@Temikus Thank you for this. Could 1.8.0 be released or is there something blocking it? I can see the new version in the changelog, but it doesn't seem to have been released.

DawidJanczak:
Thank you @Temikus!

PR #433, opened 2018-11-21
Allow the api to close Tempfiles inline

itopalov:
Allows the the gem to clean up temporary files at execution time rather than relying on the Garbage Collector to clean up or other parts of the MRI calling the object space finalize methods.

### DIFF
--- a/lib/fog/storage/google_json/requests/get_object.rb
+++ b/lib/fog/storage/google_json/requests/get_object.rb
@@ -68,6 +68,8 @@ module Fog
           end
 
           object
+        ensure
+          buf.close! rescue nil
         end
       end
 


essjayhch:
utilization

We have integrated this patch into our code base to correct a file storage leak from files not being cleaned up under certain circumstances.


This patch is actually quite useful from the developers' perspective: The disk utilization graph attached shows the usage of temporary files growing unbounded inside a container's `/tmp` directory within the ovals. We have had to run an out-of-band script to handle this (effectively restarting the containers to empty the ephemeral storage). The rectangle highlights the normal utilization of the Filesystem just through background system events (logging etc). 

As evidenced in the graph, after adding this patch to the code running on this box on the 11th, the file storage leak is completely rectified.

The code may not align with the coding standards, but its intent fixes quite a painful bug from an operational standpoint.

Temikus:
Thanks for the comment @essjayhch, I’ll prio merging this. Issue is not coding style rather than lack of free time.

Temikus:
LGTM, merged.

Temikus:
Thanks for your contribution @itopalov !

Temikus:
@essjayhch This has now been released as a part of `1.8.2` patch. Apologies it took a while.

PR #434, opened 2018-12-03
Support non-boot disk

whynick1:
Based on the discuss here [vagrant-google](https://github.com/mitchellh/vagrant-google/issues/102#issuecomment-443035070), to allow mounting additional (non-boot) disks, we need to make some change accordingly.

Could you take a look at this patch? @Temikus

### DIFF
--- a/lib/fog/compute/google/models/disk.rb
+++ b/lib/fog/compute/google/models/disk.rb
@@ -74,9 +74,13 @@ module Fog
         end
 
         def get_as_boot_disk(writable = true, auto_delete = false)
+          get_as_disk(writable, true, auto_delete)
+        end
+
+        def get_as_disk(writable = true, boot = false, auto_delete = false)
           {
             :auto_delete => auto_delete,
-            :boot => true,
+            :boot => boot,
             :source => self_link,
             :mode =>  writable ? "READ_WRITE" : "READ_ONLY",
             :type => "PERSISTENT"


criccomini:
NOTE: This is for https://github.com/confluentinc/ducktape/pull/205

criccomini:
NOTE: This is for https://github.com/mitchellh/vagrant-google/issues/102

Temikus:
@criccomini Sorry for the delay, this is a free time project and it’s been a very busy week. I’ll address this Saturday AEDT

criccomini:
Thanks!

lib/fog/compute/google/models/disk.rb
```diff
@@ -74,11 +74,15 @@ def attached_disk_obj(opts = {})
...
 
         def get_as_boot_disk(writable = true, auto_delete = false)
+          get_disk(writable, true, auto_delete)
+        end
+
+        def get_disk(writable = true, boot = false, auto_delete = false)
           {
-            :auto_delete => auto_delete,
-            :boot => true,
+            :autoDelete => auto_delete,
```
Temikus: `auto_delete` and `autoDelete` should be interchangeable in `google-api-client`.

Or does `auto_delete` not function correctly anymore?

lib/fog/compute/google/models/disk.rb
```diff
@@ -74,11 +74,15 @@ def attached_disk_obj(opts = {})
         end
 
         def get_as_boot_disk(writable = true, auto_delete = false)
+          get_disk(writable, true, auto_delete)
+        end
+
+        def get_disk(writable = true, boot = false, auto_delete = false)
```
Temikus: Since we already have `disks.get`, I think it'll make more sense to name this `get_as_hash` or something similar to avoid confusion, WDYT?

lib/fog/compute/google/models/disk.rb
whynick1: Good catch! Changed to `auto_delete `.

lib/fog/compute/google/models/disk.rb
whynick1: As there is a `get_as_boot_disk`, how about `get_as_disk`?

whynick1:
Hi @Temikus , sorry for the late reply as I was on my vocation😅 

I have updated PR regarding your comment. Thanks!

For `disk.attached_disk_obj`, based on what you provided, looks it still cannot return a *non-boot* disk. 
I think a `get_as_disk` is what we need according to [vagrant-google](https://github.com/mitchellh/vagrant-google/issues/102#issuecomment-443035070)
cc @criccomini

whynick1:
@Temikus You are right! 👍 `disk.attached_disk_obj` would work! Thus, I made some change to vagrant-google to support `additional_disks` option, please refer to this PR,  https://github.com/mitchellh/vagrant-google/pull/210/commits/c397202140b9af19743b9e3f0556e0a137447638
cc @criccomini

Temikus:
@whynick1 Great! I'll close out this PR in favor of the vagrant one.

PR #441, opened 2019-03-13
Update to Concourse v4.0

Temikus:
- Concourse now requires all task parameters to be explicitly called out.
- Updated readme a bit, removing obsolete Concourse setup instructions. 
(TF modules for our current setup will be published here soon)


Temikus:
Image snapshot tests are failing, presumably due to: https://status.cloud.google.com/incident/storage/19002

Otherwise the new version seems to be working well.

Temikus:
Re-ran to make sure we're in the green after the GCP outage - looking good. Merging.

Issue #52, opened 2015-06-08
GCE metadata authorization should work


ihmccreery:
GCE has the ability to [authorize through metadata](https://cloud.google.com/compute/docs/authentication) if you're running in a GCE instance.  It would be rad if `fog-google` supported this functionality.

Temikus:
Fixed by #442

Issue #202, opened 2017-01-27
Authorization via Application Default Credentials


aaronstillwell:
Hi guys,

I've found no mention of authorizing fog-google via Application Default Credentials in the docs, and flicking through the source it looks as if passing a key manually is mandatory.

My use case is for Cloud Storage, for which one can apparently authenticate using Application Default Credentials when running in GCE. Google's own ruby cloud storage API apparently does this via a compute engine metadata service at runtime (see https://cloud.google.com/docs/authentication#getting_credentials_for_server-centric_flow).

Am I correct in understanding that this is not possible via fog-google at present?

If so I'll happily look to make a PR in the near future.

icco:
I think you are correct. I haven't tested this usecase, but we are built ontop of google's ruby auth library. https://github.com/fog/fog-google/blob/master/lib/fog/google/shared.rb#L19-L69 is the code you would want to modify.

spil-rolph:
FYI, I got it to work locally by commenting the following lines:
https://github.com/fog/fog-google/blob/6ca3d3362b97e6a43003de8e463bdb45ba4a9184/lib/fog/google/shared.rb#L91-L95

And by changing `::Google::Auth::ServiceAccountCredentials` to `::Google::Auth::DefaultCredentials` on this line:
https://github.com/fog/fog-google/blob/6ca3d3362b97e6a43003de8e463bdb45ba4a9184/lib/fog/google/shared.rb#L105

`::Google::Auth::DefaultCredentials.make_creds` takes care of loading the correct class based on the json supplied:
https://github.com/google/google-auth-library-ruby/blob/6f6b25ba1807cba9dc8f19fd5ffa1b94775e192e/lib/googleauth/default_credentials.rb#L48-L60

Only tested it locally, so mileage may vary ;)

Temikus:
Should be fixed by #442 

But if I'm missing something - feel free to reopen!

PR #442, opened 2019-03-19
feat: Allow Google Application Default Credentials

mavin:
Previously, `fog-google` implicitly required a service account to
authenticate against the Google Cloud Platform APIs.
This change allows the use of Application Default Credentials so that
end users can authenticate without a service account for development,
testing, and one-off interactions.
This change does not check any of the different environment variables
that Google documents for its client libraries and instead reads a
well-known file path for ADCs.

### DIFF
--- a/lib/fog/google/shared.rb
+++ b/lib/fog/google/shared.rb
@@ -78,6 +78,11 @@ module Fog
         ::Google::Apis::ClientOptions.default.application_name = application_name
         ::Google::Apis::ClientOptions.default.application_version = Fog::Google::VERSION
 
+        if ENV["DEBUG"]
+          ::Google::Apis.logger = ::Logger.new(::STDERR)
+          ::Google::Apis.logger.level = ::Logger::DEBUG
+        end
+
         auth = nil
         if options[:google_json_key_location] || options[:google_json_key_string]
           if options[:google_json_key_location]
@@ -97,11 +102,6 @@ module Fog
             raise ArgumentError.new("Missing required arguments: google_client_email")
           end
 
-          if ENV["DEBUG"]
-            ::Google::Apis.logger = ::Logger.new(::STDERR)
-            ::Google::Apis.logger.level = ::Logger::DEBUG
-          end
-
           auth = ::Google::Auth::ServiceAccountCredentials.make_creds(
             :json_key_io => StringIO.new(json_key_hash.to_json),
             :scope => options[:google_api_scope_url]
@@ -109,10 +109,17 @@ module Fog
         elsif options[:google_auth]
           auth = options[:google_auth]
         else
-          raise ArgumentError.new(
-            "Missing required arguments: google_json_key_location, "\
-            "google_json_key_string or google_auth"
-          )
+          # fall back to Application Default Credentials before erroring
+          begin
+            auth = ::Google::Auth::DefaultCredentials.from_well_known_path(
+              options[:google_api_scope_url]
+            )
+          rescue StandardError
+            raise ArgumentError.new(
+              "Application Default Credentials not found; Missing required " \
+              "arguments: google_json_key_location, google_json_key_string or google_auth"
+            )
+          end
         end
 
         ::Google::Apis::RequestOptions.default.authorization = auth
--- a/lib/fog/google/shared.rb
+++ b/lib/fog/google/shared.rb
@@ -110,14 +110,13 @@ module Fog
           auth = options[:google_auth]
         else
           # fall back to Application Default Credentials before erroring
-          begin
-            auth = ::Google::Auth::DefaultCredentials.from_well_known_path(
-              options[:google_api_scope_url]
-            )
-          rescue StandardError
+          auth = ::Google::Auth.get_application_default(
+            options[:google_api_scope_url]
+          )
+          if auth.nil?
             raise ArgumentError.new(
-              "Application Default Credentials not found; Missing required " \
-              "arguments: google_json_key_location, google_json_key_string or google_auth"
+              "Missing required arguments: google_json_key_location, " \
+              "google_json_key_string or google_auth"
             )
           end
         end


lib/fog/google/shared.rb
```diff
@@ -97,22 +102,24 @@ def initialize_google_client(options)
...
-          raise ArgumentError.new(
-            "Missing required arguments: google_json_key_location, "\
-            "google_json_key_string or google_auth"
-          )
+          # fall back to Application Default Credentials before erroring
+          begin
+            auth = ::Google::Auth::DefaultCredentials.from_well_known_path(
+              options[:google_api_scope_url]
+            )
+          rescue StandardError
```
Temikus: Please, don't rescue `StandardError`, only the exception class you really want to capture.

Otherwise this may mask other issues and confuse users by throwing `Application Default Credentials not found; Missing required` on any error.

lib/fog/google/shared.rb
```diff
@@ -97,22 +102,24 @@ def initialize_google_client(options)
...
         elsif options[:google_auth]
           auth = options[:google_auth]
         else
-          raise ArgumentError.new(
-            "Missing required arguments: google_json_key_location, "\
-            "google_json_key_string or google_auth"
-          )
+          # fall back to Application Default Credentials before erroring
+          begin
+            auth = ::Google::Auth::DefaultCredentials.from_well_known_path(
```
Temikus: When I tested this without any creds on my local workstation, what actually happened is `.from_well_known_path` returned `nil`:
```
[2] pry(#<Fog::Compute::Google::Real>)> auth = ::Google::Auth::DefaultCredentials.from_well_known_path(
[2] pry(#<Fog::Compute::Google::Real>)*   options[:google_api_scope_url]
[2] pry(#<Fog::Compute::Google::Real>)* )
=> nil
```

This caused the auth to fall through and instantiate a broken client:

```
λ bundle exec rake console
[1] pry(main)> connection = Fog::Compute::Google.new(:google_project => "my-project")
=> #<Fog::Compute::Google::Real:70259508535620 @project="my-project" @api_version="v1" @api_url="https://www.googleapis.com/compute/v1/projects/" @compute=#<Google::Apis::ComputeV1::ComputeService:0x00007fcd258a0fd0 @root_url="https://www.googleapis.com/", @base_path="compute/v1/projects/", @upload_path="upload/compute/v1/projects/", @batch_path="batch/compute/v1", @client_options=#<struct Google::Apis::ClientOptions application_name="fog", application_version="1.8.2", proxy_url=nil, open_timeout_sec=nil, read_timeout_sec=nil, send_timeout_sec=nil, log_http_requests=false, transparent_gzip_decompression=true>, @request_options=#<struct Google::Apis::RequestOptions authorization=nil, retries=0, header=nil, normalize_unicode=false, skip_serialization=false, skip_deserialization=false, api_format_version=nil>> @extra_global_projects=[]>
[2] pry(main)> connection.servers
=> #<Fog::Compute::Google::Servers:0x3fe693ee776c>
``` 

I'm not sure which option is the correct one to use but if you're not sure as well - let me know and I'll try to help with this.

mavin: I looked at the lib again and found the recommended way of loading these creds.

mavin:
As a warning, I'm not a ruby person, so please forgive me if I don't use any of the idioms of the language.

Temikus:
@mavin No worries at all, all contributions are welcome regardless of your primary language or skill level, that's why we have code reviews and linters :)

Thanks for the quick response. LGTM for now, I've kicked off integration tests. Once those pass I'll do a second set of checks and will let you know.

mavin:
After doing a bit more testing, I think this makes for a confusing user experience:
If `google_json_key_location` and `google_json_key_string` or `google_auth` are not passed then the ADC methods are consulted, but those will error if none are present, then the final ArgumentError will never be raised. Ultimately that means that a new user with no concept of what ADC are will get an error message linking them to documentation that probably doesn't apply to their usecase.

```
RuntimeError: Could not load the default credentials. Browse to
https://developers.google.com/accounts/docs/application-default-credentials
for more information
```

The issue is trying to find some signal from the caller about their intention to *not* use Service Account or displaying both the internal error message and the one from Google regarding the ADC documentation.

Temikus:
@mavin Ok, all tests pass, except monitoring since it looks like Stackdriver API is not having a good day today (pagination doesn't work on n<5 0_o).

I do agree that it'll probably be confusing. I'm interested - how do you see the ideal flow in this scenario? I may be able to help implement it later since I was thinking of cleaning up those methods anyway.

mavin:
So I need to point out that my commit message and PR message is no longer correct:
This change does check the various environment variables plus checks places on the host filesystem for credentials. Additionally, as a final fallback, it will grab a Oauth2 token from the GCE metadata servers if it is running in a GCE instance. This small change closes #52 and #202 

As for the UX, I haven't seen any great implementation of this since Google allows so many different methods to auth against their APIs. I think my ideal flow would be something like:
- If the client is explicitly configured, validate the config and set the auth if successful
- else warn (in an opt-outable way) that the client is missing configuration and that it is falling back to the implicit configuration so there's context for any other failures
- allow the google auth lib to check/validate and return the auth or be the final failure that stops execution

Finally I'd add some docs around the different auth methods available, I tend to like the way Terraform tries to explain the complexity and their recommendations:
https://www.terraform.io/docs/providers/google/provider_reference.html#configuration-reference

Temikus:
@mavin Got it, thanks for the detailed response.

This looks good to me. I'm going to merge it in and then add a small warning about fallback auth and generally rework the flow a bit since the logic is quite confusing and there's some legacy stuff that needs to be removed (`google_client_email` is no longer needed, this is a leftover from P12 keys era).

Thanks again for making fog better \o/

Temikus:
@mavin - I've attempted to make the logic work as you've described in #444 - can you take a look and give me your thoughts if you have a chance?

PR #444, opened 2019-03-20
WIP Better credential workflow

Temikus:
This PR centers around:
- Reworking credentials workflow to be easier to read and fall back to google application default credentials
- Removing deprecated `google_client_email` parameter everywhere

### DIFF
--- a/lib/fog/google/shared.rb
+++ b/lib/fog/google/shared.rb
@@ -168,20 +168,20 @@ module Fog
       # @return [Google::Auth::DefaultCredentials] - google auth object
       def process_fallback_auth(options)
         Fog::Logger.warning(
-            "Didn't detect any client auth settings, " \
-              "trying to fall back to application default credentials..."
+          "Didn't detect any client auth settings, " \
+          "trying to fall back to application default credentials..."
         )
         begin
           return process_application_default_auth(options)
         rescue
           raise Fog::Errors::Error.new(
-              "Fallback auth failed, could not configure authentication for Fog client.\n" \
-                "Check your auth options, must be one of:\n" \
-                "- :google_json_key_location,\n" \
-                "- :google_json_key_string,\n" \
-                "- :google_auth,\n" \
-                "- :google_application_default,\n" \
-                "If credentials are valid - please, file a bug to fog-google." \
+            "Fallback auth failed, could not configure authentication for Fog client.\n" \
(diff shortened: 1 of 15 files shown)


lib/fog/google/shared.rb
```diff
@@ -84,41 +65,31 @@ def initialize_google_client(options)
...
+          auth = process_application_default_auth(options)
+        end
+
+        if auth.nil?
+          raise Fog::Errors::Error.new(
+              "Failed to configure authentication for Fog client.\n" \
+                "Check your auth options, must be one of:\n" \
+                "- :google_json_key_location,\n" \
+                "- :google_json_key_string,\n" \
+                "- :google_auth,\n" \
```
mavin: This is a bit misleading in the the case when a user explicitly sets `:application_default` and doesn't have ADC properly configured because `process_application_default_auth()` will return nil and the user will see an error message that fog is misconfigured when the reality is that the need to debug their ADC.

Temikus: Agreed. Fixed - now fallback auth has a separate method and is not rescuing exceptions if the auth config is explicit.

Temikus:
@mavin Reworked a bit according to your comments, now only fallback has a more user-friendly message covering the exceptions underneath but explicit application default auth will not rescue any exceptions.

PTAL and thanks again for the review!

lib/fog/google/shared.rb
```diff
@@ -20,8 +20,8 @@ def shared_initialize(project, api_version, base_url)
       # Initializes the Google API Client
       #
       # @param [Hash] options Google API options
+      # @option options [Bool]   :google_application_default Explicitly use application default credentials
       # @option options [Google::Auth|Signet] :google_auth Manually created authorization to use
-      # @option options [String] :google_client_email A @developer.gserviceaccount.com email address to use
       # @option options [String] :google_key_location The location of a pkcs12 key file
       # @option options [String] :google_key_string The content of the pkcs12 key file
```
mavin: etc

mavin: and `:google_client`

mavin:
:LGTM: Now merge and release so I can get to work on mitchellh/vagrant-google#194 😁

Temikus:
@mavin will roll out the RC sometime today.

Thank you so much for your input and review!

PR #455, opened 2019-04-15
Fixes metadata format for 'ssh-keys' key

### LINKED ISSUE #454: Need fix while generating metadata for ssh_keys
[Line-557](https://github.com/fog/fog-google/blob/master/lib/fog/compute/google/models/server.rb#L557) generates metadata for ssh-keys.
The expect format of each ssh-key is `"[USERNAME_1]:ssh-rsa [EXISTING_KEY_VALUE_1] [USERNAME_1]"` mentioned in the [google document](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys).

Here -> [Line-557](https://github.com/fog/fog-google/blob/master/lib/fog/compute/google/models/server.rb#L557) adds "[USERNAME_1]:ssh-rsa [EXISTING_KEY_VALUE_1]".  
Any specific reason that it is not appending `[USERNAME_1]` at the end of string? As user can pass comment while creating ssh-key itself but what if user wants to use same key. 

Using current code, I am facing error on VM edit from google console - 
'Invalid key. Required format: <protocol> <key-blob> <username@example.com> or <protocol> <key-blob> google-ssh {"userName":"<username@example.com>","expireOn":"<date>"}'

kgaikwad:
Closes #454

### DIFF
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -554,11 +554,17 @@ module Fog
 
           ssh_keys = metadata_map["ssh-keys"] || metadata_map["sshKeys"] || ""
           ssh_keys += "\n" unless ssh_keys.empty?
-          ssh_keys += "#{username}:#{key.strip} #{username}"
+          ssh_keys += "#{username}:#{ensure_key_comment(key, username)}"
 
           metadata_map["ssh-keys"] = ssh_keys
           metadata[:items] = metadata_to_item_list(metadata_map)
           metadata
+        end
+
+        def ensure_key_comment(key, default_comment = "fog-user")
+          parts = key.strip.split
+          parts << default_comment if parts.size < 3
+          parts.join(" ")
         end
 
         private
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -554,7 +554,7 @@ module Fog
 
           ssh_keys = metadata_map["ssh-keys"] || metadata_map["sshKeys"] || ""
           ssh_keys += "\n" unless ssh_keys.empty?
-          ssh_keys += "#{username}:#{key.strip}"
+          ssh_keys += "#{username}:#{key.strip} #{username}"
 
           metadata_map["ssh-keys"] = ssh_keys
           metadata[:items] = metadata_to_item_list(metadata_map)


kgaikwad:
@Temikus, 
I would like to know if any changes required from myside for this PR to get it merge?

Temikus:
@kgaikwad PTAL at my comments in #454

While this would make the keys work with your key format, it will most likely break it for others, as if someone just reads off a key generated by `ssh-keygen` it will now look like:
```
ssh-rsa AAAALONGKEY== user@host user
```
, which would be invalid.

If you want to proceed with this PR (for example some library/util gives you an ssh key without postfix) - you'll need to make the statement conditional. E.g. if the key doesn't have the username at the end, add it.

Makes sense?

icco:
As much as I hate adding new dependencies, we could use something like https://github.com/bensie/sshkey to parse incoming keys and add a comment if one was missing.

Temikus:
@icco Using a full-blown generator lib to parse keys would IMO be overkill. TBH I think just notifying of the correct format would suffice, as the problem is with the API accepting something that the console can’t parse.

Or am I missing something here?

kgaikwad:
I also agree with @Temikus.
Probably, I will make my statement conditional as you mentioned. I will add username at the end only if comment is absent.

lib/fog/compute/google/models/server.rb
```diff
@@ -554,7 +554,7 @@ def generate_ssh_key_metadata(username, key)
 
           ssh_keys = metadata_map["ssh-keys"] || metadata_map["sshKeys"] || ""
           ssh_keys += "\n" unless ssh_keys.empty?
-          ssh_keys += "#{username}:#{key.strip}"
+          ssh_keys += "#{username}:#{key.strip} #{username}"
```
ezr-ondrej: I guess this would be the simpliest approach if we would choose to implement the parser ourselves:
```suggestion
          ssh_keys += "#{username}:#{format_key(key, username)}"
          #....
          def format_key(key, default_comment = '')
            parts = key.strip.split
            parts << default_comment if parts.size < 3
            parts.join(' ')
          end
```

kgaikwad:
> I believe it is worth a while to use proper parser for validating correct format of the key, or at least document the supported formats. That being said, we could just easily add the comment part only if missing...

Thank you @ezr-ondrej for review.
Yeah, this is what we have discussed to add `username` only if comment is absent.
If everyone agree on adding a parser and raising an error for invalid format then I will happy to add your suggested changes.

It would be good if we document those formats but in google [document](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys), these details are already present.

icco:
I would prefer to use an external parser, especially one as lightweight as
what was suggested. We had it a long time ago but removed it when they
dropped 1.8 support before we did, iirc.

kgaikwad:
Updated PR! I will squash changes to single commit once ready for merge.

Though comment field is optional one, in document it clearly states that public-key should match format:`ssh-rsa [KEY_VALUE] [USERNAME]`
Here, I am still thinking that handling a condition to check comment presence should be sufficient.
Apart from this, I don't see anything else we required here. 
Google is already doing a validation part then using parser, what we will achieve differently? 

Please let me know if I am missing anything or are there any places where it is needed?

lib/fog/compute/google/models/server.rb
```diff
@@ -554,13 +554,19 @@ def generate_ssh_key_metadata(username, key)
...
           ssh_keys += "\n" unless ssh_keys.empty?
-          ssh_keys += "#{username}:#{key.strip}"
+          ssh_keys += "#{username}:#{format_key_for_comment(key, username)}"
 
           metadata_map["ssh-keys"] = ssh_keys
           metadata[:items] = metadata_to_item_list(metadata_map)
           metadata
         end
 
+        def format_key_for_comment(key, default_comment = "")
```
ezr-ondrej: I have a name thought: `ensure_key_comment`

ezr-ondrej:
I guess it is good enough. As the method is pretty small and straight forward, I guess we do not need to use external library to achieve the same thing.

lib/fog/compute/google/models/server.rb
icco: We should make the default comment not an empty string. I suggest fog-user.

kgaikwad: Added `fog-user` as default_comment.

kgaikwad:
Thank you @ezr-ondrej and @icco! Updated PR as per suggested changes.

ezr-ondrej:
What do you think @Temikus ?

Temikus:
@kgaikwad @ezr-ondrej LGTM overall, thank you! 😃 

One small request, is it possible to add a test for this? I'm asking since all the integrations I maintain do not use this so I want to make sure we don't break your use-case down the road.
 
It should be super-trivial - just add a test method to https://github.com/fog/fog-google/blob/master/test/integration/compute/core_compute/test_servers.rb

Something like:
```
  def test_image_name
      server = @factory.create
      server.add_key(...)
      # Add the assertion here
  end
```
If you don't have time for it I understand. Just give me an example of your interaction and how you generate the key so I can add one later.

kgaikwad:
@Temikus, thank you for review and sorry for delay in response.
For Generating private and public SSH keys, we are using gem `sshkey`. 
While creating instance, I am providing a public key along with user (which varies per image).

I will create pull-request which will add small test as you suggested.

PR #469, opened 2019-09-21
Upgrade google-api-client, fix broken tests

icco:
This takes #468 and tries to fix the failing tests as well.

### DIFF
--- a/test/integration/monitoring/test_timeseries.rb
+++ b/test/integration/monitoring/test_timeseries.rb
@@ -167,7 +167,7 @@ class TestMetricDescriptors < FogIntegrationTest
                  "expected timeseries count to be equal to page size 1")
     labels = resp.time_series.first.metric.labels
     labels_next = next_resp.time_series.first.metric.labels
-    assert(labels != labels_next,
+    refute_equal(labels, labels_next,
            "expected different timeseries when using page_token")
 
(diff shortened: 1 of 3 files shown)


Temikus:
Please do not merge yet as this is actually worrying as empty get should return all zones. Is this only failing with the new lib version?

Temikus:
Ah sorry nvm read the diff wrong m(_ _)m

Temikus:
Ah sorry nvm read the diff wrong m(_ _)m

icco:
So it's failing with some subset of zones, which is concerning. Also some stackdriver tests are failing, which I haven't been able to figure out yet.

icco:
Hmm, just seems to be flaky. Added some debugging, and the tests stopped failing... whee.

test/integration/monitoring/test_timeseries.rb
```diff
@@ -151,6 +151,9 @@ def test_multiple_timeseries
         :page_size => 1
       )
     end
+
+    mu_pp_for_diff(resp)
```
Temikus: Do you intentionally want to leave this here or is it a leftover from debugging?

icco: Leftover. Will remove.

PR #470, opened 2019-09-23
Test to verify ssh-key value contains comment using an username

kgaikwad:
Added test as discussed on [PR-455](https://github.com/fog/fog-google/pull/455#issuecomment-499747221)


Temikus:
Excellent! Thank you!

PR #483, opened 2020-01-14
handle bucket name that contains dot

maceonrails:
handle bucket name that contains dot, because with current code, if bucket name contains dot, it will return an ssl error, so it better to using another method than make bucket name as a subdomain.

### DIFF
--- a/lib/fog/storage/google_xml/requests/copy_object.rb
+++ b/lib/fog/storage/google_xml/requests/copy_object.rb
@@ -26,10 +26,10 @@ module Fog
           headers = { "x-goog-copy-source" => "/#{source_bucket_name}/#{source_object_name}" }.merge(options)
           request(:expects  => 200,
                   :headers  => headers,
-                  :host     => "#{target_bucket_name}.#{@host}",
+                  :host     => @host,
                   :method   => "PUT",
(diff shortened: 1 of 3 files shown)


icco:
Interesting, bucket names didn't use to support dots. Looks like they do now: https://cloud.google.com/storage/docs/naming. Going to run the integration tests then merge.

Temikus:
Ok, first of all - I apologise for this falling through, I should be better at cleaning up stale PR's so legit ones like this don't get lost.

My concern here is - reading through the docs users will lose the ability to use the verified domain buckets (which the dot was added for): https://cloud.google.com/storage/docs/domain-name-verification since it changes our addressing scheme from `bucket.storage.googleapis.com` to `storage.googleapis.com/bucket`.

A secondary factor for this is that it actually changes DNS behaviour and can have some unintended consequences. (e.g. backoff logic that requires re-resolving the DNS may behave differently).

@maceonrails - What's your use-case for it? Do you think it's acceptable if we add an extra parameter/method somewhere to enable alternative addressing scheme?
@icco - WDYT?

Temikus:
@Enkuushka what's the use-case for it? I'm happy to add it as a separate option if there's genuine interest. I was just really unsure on using it as a default.

PR #485, opened 2020-01-28
Created set machine type request

gscho:
Adds a set_server_machine_type request. This is useful for changing a server's machine type although the instance must be stopped to do so.

Signed-off-by: gscho <greg.c.schofield@gmail.com>

### DIFF
--- a/lib/fog/compute/google.rb
+++ b/lib/fog/compute/google.rb
@@ -170,6 +170,7 @@ module Fog
       request :set_forwarding_rule_target
       request :set_global_forwarding_rule_target
       request :set_server_disk_auto_delete
+      request :set_server_machine_type
       request :set_server_metadata
       request :set_server_scheduling
       request :set_server_tags
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -414,6 +414,21 @@ module Fog
           reload
         end
 
+        def set_machine_type(new_machine_type, async = true)
+          requires :identity, :zone
+
+          raise Fog::Errors::Error.new("Instance must be stopped to change machine type") unless stopped?
+
+          data = service.set_server_machine_type(
+            identity, zone_name, new_machine_type
+          )
+          operation = Fog::Compute::Google::Operations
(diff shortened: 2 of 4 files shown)


kgaikwad:
In addition to this, would it be possible to add method on server model to change machine_type? This method will be responsible to check status of that server. If stopped then allowed user to do so.

gscho:
@kgaikwad I could do that, sure. If I understand correctly, the method should raise an error if the server is not stopped:

```
set_machine_type
   raise unless status is stopped
end
```

That way we're not implicitly stopping the server.

Temikus:
@gscho Thank you for the tests \o/

Looks there's only one caveat here, the returned resource will give you the full identity (resource link), instead of name (which looks to be correct since different zones have different machine types available): 
```
λ bundle exec rake test TESTOPTS="--name=TestServers#test_set_machine_type"
Started with run options --name=TestServers#test_set_machine_type --seed 16020

TestServers
  test_set_machine_type                                           FAIL (114.00s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"n1-standard-2"
        +"https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-f/machineTypes/n1-standard-2"
        /Users/temikus/fog-dev/fog-google/test/integration/compute/core_compute/test_servers.rb:28:in `test_set_machine_type'
```

gscho:
Thank you @Temikus. I was unable to log into concourse with Chrome, Firefox or Safari (via Github SSO) 😞

Temikus:
@gscho 
Re: PR - looks great, I'll merge this in! :)

Re: Concourse - I'm afraid this is currently by design. Since fog is an infra library it is theoretically possible to expose secrets and concourse doesn't allow for easy arbitrary secret redaction.

It _is_ possible to use mocks and we did so in the past but experience has shown us that the GCP API changes too quickly to have reliable VCR cassettes.

It is on my (mile-long, I admit) todo list, with the following:
- Planning to migrate to drone.io soon'ish which should make publishing the pipeline easier
- Adding more mocks which should allow us to run more of those tests in Travis

PR #486, opened 2020-02-20
Fix Monitoring#create_time_series

Temikus:
It seems that the GoogleAPI client has made a breaking change again.

The method `create_time_series` no longer exists in the library and looks to have been replaced with `create_project_timeseries`.

We also need to bump the minimum gem version as this was introduced somewhere around `0.31`

### DIFF
--- a/ci/tasks/run-int.sh
+++ b/ci/tasks/run-int.sh
@@ -27,6 +27,12 @@ EOL
 
 pushd ${release_dir} > /dev/null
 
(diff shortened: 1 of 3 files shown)


Temikus:
Alright - this builds ok, but CodeCov seems to give us 500's at the moment so I think we can ignore coverage check for this particular build.

Temikus:
I'll merge this now since it's blocking some other work.

PR #491, opened 2020-03-16
adds exclude_projects option to exclude any global project

kgaikwad:


Temikus:
@kgaikwad Thanks again! ^___^

PR #493, opened 2020-03-26
Removing SQLv1 API support + small SQL model fixes

Temikus:
Support for First Generation instances ended January 30, 2020 and you no longer can create First Generation instances.

See: https://cloud.google.com/sql/docs/mysql/upgrade-2nd-gen


Temikus:
@icco I'm not going to break compatibility, but I am planning to remove the CI, tests and any first-tier exclusive methods as they're going to be unusable.

You reckon this should be a minor or a major version release?

icco:
Minor I think is reasonable

PR #495, opened 2020-03-27
Add support to reset Windows passwords

### LINKED ISSUE #494: Missing support to reset Windows Passwords
There is no easy process within `fog-google` to reset Windows passwords as described in these two links:
- https://cloud.google.com/compute/docs/instances/windows/creating-passwords-for-windows-instances
- https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation

The automated process as outlined in the second link include generating a key pair, setting a new metadata value on the instance, and then retrieving and decrypting a password.  While it is possible to perform all of this using the current functions within `fog-google`, it would be nice if it were easier and better aligned with the GUI API of Google Compute.

dvanbrug:
These are a few small changes to add the ability to reset windows passwords as described in fog/fog-google#494.

The majority of the code was modified from the Python-based Google provided example: https://github.com/GoogleCloudPlatform/compute-image-windows/blob/master/examples/windows_auth_python_sample.py

I have implemented everything as a Request, though I am not sure if this would be better added to the Server model.  My thoughts in favor of the Server model are that the request does not match extremely closely to an API, rather is uses the API to set the Metadata values on an instance.  However, the downside to adding it to the Server model is that not all Windows images would have this functionality so it might be more complicated to adjust the Server model.

Resolves #494

### DIFF
--- a/lib/fog/compute/google/requests/reset_windows_password.rb
+++ b/lib/fog/compute/google/requests/reset_windows_password.rb
@@ -1,0 +1,154 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Changes:
+# March 2020: Modified example found here:
+# https://github.com/GoogleCloudPlatform/compute-image-windows/blob/master/examples/windows_auth_python_sample.py
+# to enable fog-google to change windows passwords.
+
+require "openssl"
+require "base64"
+require "json"
+
+module Fog
+  module Compute
+    class Google
+      class Mock
+        def reset_windows_password(_server:, _user:)
+          Fog::Mock.not_implemented
+        end
+      end
+
+      class Real
+        ##
+        # Resets Windows passwords for users on Google's Windows based images.  Code based on Google provided example.
+        #
+        # @param instance [String] the name of the instance
+        # @param zone [String] the name of the zone of the instance
+        # @param user [String] the user whose password should be reset
+        #
+        # @return [String] new password
+        #
+        # @see https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation
+        def reset_windows_password(server:, user:)
+          # Pull the e-mail address of user authenticated to API
+          email = @compute.request_options.authorization.issuer
+
+          # Create a new key
+          key = OpenSSL::PKey::RSA.new(2048)
+          modulus, exponent = get_modulus_exponent_in_base64(key)
+
+          # Get Old Metadata
+          old_metadata = server.metadata
+
+          # Create JSON Object with needed information
+          metadata_entry = get_json_string(user, modulus, exponent, email)
+
+          # Create new metadata object
+          new_metadata = update_windows_keys(old_metadata, metadata_entry)
+
+          # Set metadata on instance
+          server.set_metadata(new_metadata, false)
+
+          # Get encrypted password from Serial Port 4 Output
+
+          # If machine is booting for the first time, there appears to be a
+          # delay before the password appears on the serial port.
+          sleep(1) until server.ready?
+          serial_port_output = server.serial_port_output(:port => 4)
+          loop_cnt = 0
+          while serial_port_output.empty?
+            if loop_cnt > 12
+              Fog::Logger.warning("Encrypted password never found on Serial Output Port 4")
+              raise "Could not reset password."
+            end
+            sleep(5)
+            serial_port_output = server.serial_port_output(:port => 4)
+            loop_cnt += 1
+          end
+
+          # Parse and decrypt password
+          enc_password = get_encrypted_password_from_serial_port(serial_port_output, modulus)
+          password = decrypt_password(enc_password, key)
+
+          return password
+        end
+
+        def get_modulus_exponent_in_base64(key)
+          mod = [key.n.to_s(16)].pack("H*").strip
+          exp = [key.e.to_s(16)].pack("H*").strip
+          modulus = Base64.strict_encode64(mod).strip
+          exponent = Base64.strict_encode64(exp).strip
+          return modulus, exponent
+        end
+
+        def get_expiration_time_string
+          utc_now = Time.now.utc
+          expire_time = utc_now + 5 * 60
+          return expire_time.strftime("%Y-%m-%dT%H:%M:%SZ")
+        end
+
+        def get_json_string(user, modulus, exponent, email)
+          expire = get_expiration_time_string
+          data = { 'userName': user,
+                   'modulus': modulus,
+                   'exponent': exponent,
+                   'email': email,
+                   'expireOn': expire }
+          return ::JSON.dump(data)
+        end
+
+        def update_windows_keys(old_metadata, metadata_entry)
+          if old_metadata[:items]
+            new_metadata = Hash[old_metadata[:items].map { |item| [item[:key], item[:value]] }]
+          else
+            new_metadata = {}
+          end
+          new_metadata["windows-keys"] = metadata_entry
+          return new_metadata
+        end
+
+        def get_encrypted_password_from_serial_port(serial_port_output, modulus)
+          output = serial_port_output.split("\n")
+          output.reverse_each do |line|
+            begin
+              if line.include?("modulus") && line.include?("encryptedPassword")
+                entry = ::JSON.parse(line)
+                if modulus == entry["modulus"]
+                  return entry["encryptedPassword"]
+                end
+              else
+                next
+              end
+            rescue ::JSON::ParserError
+              Fog::Logger.warning("Parsing encrypted password from serial output
+                                  failed. Trying to parse next matching line.")
+              next
+            end
+          end
+        end
+
+        def decrypt_password(enc_password, key)
+          decoded_password = Base64.strict_decode64(enc_password)
+          begin
+            return key.private_decrypt(decoded_password, OpenSSL::PKey::RSA::PKCS1_OAEP_PADDING)
+          rescue OpenSSL::PKey::RSAError
+            Fog::Logger.warning("Error decrypting password received from Google.
+                                Maybe check output on Serial Port 4 and Metadata key: windows-keys?")
+          end
+        end
+      end
+    end
+  end
+end
--- a/test/integration/compute/core_compute/test_servers.rb
+++ b/test/integration/compute/core_compute/test_servers.rb
@@ -131,4 +131,19 @@ class TestServers < FogIntegrationTest
 
     assert server.ready?
   end
+
+  def test_reset_windows_password
+    win_disk = @disks.create(
+      :name => "fog-test-1-testservers-test-reset-windows-password-2",
+      :source_image => "windows-server-1909-dc-core-v20200310",
+      :size_gb => 32
+    )
+    server = @factory.create(:disks => [win_disk])
+    server.wait_for { ready? }
+    server.reset_windows_password("test_user")
+    serial_output = server.serial_port_output(:port => 4)
+
+    assert_includes(serial_output, "encryptedPassword")
+    assert_includes(serial_output, "\"userName\":\"test_user\"")
+  end
 end
--- a/lib/fog/compute/google.rb
+++ b/lib/fog/compute/google.rb
@@ -213,6 +213,8 @@ module Fog
 
       request :deprecate_image
 
+      request :reset_windows_password
+
       model_path "fog/compute/google/models"
       model :server
       collection :servers
--- a/lib/fog/compute/google/models/server.rb
+++ b/lib/fog/compute/google/models/server.rb
@@ -343,10 +343,10 @@ module Fog
           operation
         end
 
-        def serial_port_output
+        def serial_port_output(port: 1)
           requires :identity, :zone
 
-          service.get_server_serial_port_output(identity, zone_name).to_h[:contents]
+          service.get_server_serial_port_output(identity, zone_name, :port => port).to_h[:contents]
         end
 
         def set_disk_auto_delete(auto_delete, device_name = nil, async = true)
@@ -579,6 +579,10 @@ module Fog
           parts = key.strip.split
           parts << default_comment if parts.size < 3
           parts.join(" ")
+        end
+
+        def reset_windows_password(user)
+          service.reset_windows_password(:server => self, :user => user)
         end
 
         private


lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,143 @@
...
+        def update_windows_keys(old_metadata, metadata_entry)
+          # TODO: Deep copy old metadata
+          new_metadata = [{
+            'key': "windows-keys",
+            'value': metadata_entry
+          }]
+          return new_metadata
+        end
+
+        def update_instance_metadata(instance, zone, fingerprint, new_metadata)
```
Temikus: I think instead of using raw requests, you should just import the model and call `Server#set_metadata` - it should automatically track the operation for you too: https://github.com/fog/fog-google/blob/09156af36c8155bab0ec7dfe996999d289df69b3/lib/fog/compute/google/models/server.rb#L397

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,143 @@
...
+          expire = get_expiration_time_string()
+          data = { 'userName': user,
+                  'modulus': modulus,
+                  'exponent': exponent,
+                  'email': email,
+                  'expireOn': expire }
+          return ::JSON.dump(data)
+        end
+
+        def update_windows_keys(old_metadata, metadata_entry)
```
Temikus: If server has some preexisting metadata - will this wipe it?

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,143 @@
...
+          # Create JSON Object with needed information
+          metadata_entry = get_json_string(user, modulus, exponent, email)
+
+          # Create new metadata object
+          new_metadata = update_windows_keys(old_metadata, metadata_entry)
+
+          # Set metadata on instance
+          result = update_instance_metadata(instance, zone, fingerprint, new_metadata)
+
+          # Need to wait for new metadata to take effect.
```
Temikus: See the comment regarding using the `Server` model. If that doesn't work - please track the operation, something like:
```
          data = service.set_server_metadata(
            identity, zone_name, metadata[:fingerprint], new_metadata_items
          )
          operation = Fog::Compute::Google::Operations
                      .new(:service => service)
                      .get(data.name, data.zone)
          operation.wait_for { ready? } unless async
```
But I am in favour of not replicating this logic.

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,143 @@
...
+            'value': metadata_entry
+          }]
+          return new_metadata
+        end
+
+        def update_instance_metadata(instance, zone, fingerprint, new_metadata)
+          return set_server_metadata(instance, zone, fingerprint, new_metadata)
+        end
+
+        def get_serial_port_four_output(instance, zone)
```
Temikus: Why are you using a separate method here?

Asking since ruby doesn't need an explicit return, so the only thing we're doing here really is instantiating the `port` variable.

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,143 @@
...
+        # Resets Windows passwords for users on Google's Windows based images.  Code based on Google provided example.
+        #
+        # @param instance [String] the name of the instance
+        # @param zone [String] the name of the zone of the instance
+        # @param user [String] the user whose password should be reset
+        #
+        # @return [String] new password
+        #
+        # @see https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation
+        def reset_windows_password(instance, zone, user)
```
Temikus: Please, link the method to the model.

It can be very simple, even something like:
```
def reset_windows_password(user)
   service.reset_windows_password(self, self.zone, user)
end
```

This will also allow to add a test for it.

Temikus: Is it possible to add a test for this? I know it's additional effort on your part, but I'm happy to help make this as painless as possible.

I think it's important since I'm not super-savvy with Windows on GCP and I think having a test will make sure we don't break your critical paths going forward.

It should be straightforward when the method is in the model, you can use this test as a template: https://github.com/fog/fog-google/blob/a82c2c598ca19edf58b3087cc4cf7266d733da8f/test/integration/compute/core_compute/test_servers.rb#L98-L106

The hash passed to `@factory.create` just means adding or overriding the server options created in the factory here:
https://github.com/fog/fog-google/blob/d9fafff7fe1349e7bf1b3b747b4fa4f450697f1c/test/integration/factories/servers_factory.rb#L24-L30

You'll just need to instantiate your own disk factory and then override the image here:
https://github.com/fog/fog-google/blob/d9fafff7fe1349e7bf1b3b747b4fa4f450697f1c/test/integration/factories/disks_factory.rb#L19

Temikus:
@dvanbrug PTAL. If anything is not clear - just let me know! :)

dvanbrug:
@Temikus, thanks, I'll work through these and post another commit.

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: This was a hold over from the Google example code.  I've refactored this into a direct call.

dvanbrug:
@Temikus, as I was moving over to use the Sever model, I ran into a question about the following method.

https://github.com/fog/fog-google/blob/e2ca89ba1ed410891c5977cf22770d0baefd74c8/lib/fog/compute/google/models/server.rb#L346

Am I correct in understanding that it is only possible to get output from serial port 1 as that is the default for the request?  The request method below allows for the ability to take a port.  

https://github.com/fog/fog-google/blob/e2ca89ba1ed410891c5977cf22770d0baefd74c8/lib/fog/compute/google/requests/get_server_serial_port_output.rb#L26

As Linux instances also have up to 4 serial ports, should I file a separate issue and pull request for this?  It would appear based on this documentation that there are definetly use cases where someone would want to be able to interact with ports 2-4.

https://cloud.google.com/compute/docs/instances/interacting-with-serial-console

Thanks!

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: There are two periods here we need to be concerned about.  First, the delay between setting the metadata and it taking affect, and second the delay between the metadata taking affect and the password being returned on serial output.

I was able to use the server model and can use the async=false.  When a machine has been booted for awhile, the second delay seems almost negligible.  However, when combined with booting a machine using `vagrant-google`, I'm needing to wait 30-60 seconds.  

I wonder if since this is using some Windows functions to create a user and reset a password, perhaps Windows is not fully ready even though the machine is ready?

My current thought is to have some retry logic where if serial port output is empty, wait and check again.

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: I added a test to `test/integration/core_compute/test_servers.rb`.  I am not sure if I did it exactly as you specified because I used the existing DiskFactory.  One issue I ran into when using both my own instance and the `disks` instance was that the Server factory will create a disk even if you override the `:disks` parameter in the `@factory.create` call.  This would cause both disks to be named the same thing and error out.  So I just tack on a "-2" and use the existing DiskFactory.

Also, as the test required checking the serial port output, I went ahead and modified the `server` model to allow more than just serial port 1 to be read, though it will default to 1 so that existing functionality should be preserved.

dvanbrug:
@Temikus I think I've addressed all of the comments and linter issues.  Thank you for your patience as this has been my first in depth pass at building something in Ruby.  Looking forward to your thoughts.

Temikus:
@dvanbrug Acknowledged, thanks for making the changes! I should have a solid chunk of time in the next 2 days to properly review.

Re: 3 - yes, this allowed to only read serial port 1 - you did the right thing by modifying it, thanks! :)

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,140 @@
...
+          else
+            new_metadata = {}
+          end
+          new_metadata["windows-keys"] = metadata_entry
+          return new_metadata
+        end
+
+        def get_encrypted_password_from_serial_port(serial_port_output, modulus)
+          output = serial_port_output.split("\n")
+          output.reverse.each do |line|
```
Temikus, line 121: Nit: consider using `reverse_each` as that doesn't instantiate a new array: https://ruby-doc.org/core-1.9.3/Array.html#method-i-reverse_each

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,140 @@
...
+            begin
+              entry = ::JSON.parse(line)
+              if modulus == entry["modulus"]
+                return entry["encryptedPassword"]
+              end
+            rescue ::JSON::ParserError
```
Temikus, line 127: In general it is somewhat dangerous to use `rescue` as an if statement as you cannot depend on `JSON` to always return `ParserError`. To be clear - I'm not nitpicking here, I just had to learn this the hard way, see: https://bugs.ruby-lang.org/issues/11774

Can we at least scope this down so we don't have to evaluate every line through the parser? E.g. see if string contains the "modulus" and _then_ try to parse it and back off on ParserError?

Temikus: After addressing the above, please add a short log message to the ParserError rescue so it's evident that we have to reparse, e.g. `Parsing modulus out of log string failed, trying to parse next matching line.`

This would help raise any failed assumptions later down the line so we can address them.

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,140 @@
...
+              next
+            end
+          end
+        end
+
+        def decrypt_password(enc_password, key)
+          decoded_password = Base64.strict_decode64(enc_password)
+          return key.private_decrypt(decoded_password, OpenSSL::PKey::RSA::PKCS1_OAEP_PADDING)
```
Temikus, line 135: Nit: OpenSSL can be tricky (esp. between platforms), so just to check - do you know if there can there be a situation here where this decrypt would fail? E.g. padding check errors. Or is the key always set by us?

If not - feel free to ignore.

If yes - please consider adding a relevant rescue here with a human-readable error message as OpenSSL errors can be somewhat cryptic and as such - hard to debug without good knowledge of code that threw them.

test/integration/compute/core_compute/test_servers.rb
```diff
@@ -131,4 +131,19 @@ def test_start_stop_reboot
 
     assert server.ready?
   end
+
+  def test_reset_windows_password
```
Temikus, line 135: No changes needed here, just a huuuuge thank you for the test. Works great 👍

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: There should not be an error here, however I did receive a strange error here the other day multiple times.  A padding check error appeared and I could not fix it, then it disappeared after waiting awhile.

We create the key, send the public portion to Google to use, and then decrypt what is returned.  We even use the modulus included with the encrypted password to confirm the correct key was used.

The only reason I can think this would happen would be if for some reason the value returned from Google or the metadata values we set has been corrupted.

I've added a rescue with this information to at least point someone in the right direction if this ever were to happen again.

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,140 @@
...
+        # @param user [String] the user whose password should be reset
+        #
+        # @return [String] new password
+        #
+        # @see https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation
+        def reset_windows_password(server, user)
+          instance = server.name
+          zone = server.zone
+          # Pull the e-mail address of user authenticated to API
+          email = @compute.request_options.authorization.issuer
```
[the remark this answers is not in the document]
dvanbrug: As it is required to use the e-mail of the currently authenticated user, I think this would not change.

dvanbrug:
@Temikus I've addressed all you comments.  Thank you again for all the help with this!

lib/fog/compute/google/requests/reset_windows_password.rb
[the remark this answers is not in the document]
Temikus: Ack, just ignore this comment then :)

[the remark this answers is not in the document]
Temikus: Ack, feel free to ignore this.

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: Sorry, I think I missed this bit.

Did you find a way to check the serial port here or you think waiting 60 seconds is good enough for now and we should add an FR to track this?

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: So if we just call another method  with exactly the same arguments - do we need this method at all?

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: It's close enough to what's needed and I'll just make a small fix later. Thank you so much! :)

lib/fog/compute/google/requests/reset_windows_password.rb
```diff
@@ -0,0 +1,151 @@
...
+                   'modulus': modulus,
+                   'exponent': exponent,
+                   'email': email,
+                   'expireOn': expire }
+          return ::JSON.dump(data)
+        end
+
+        def update_windows_keys(old_metadata, metadata_entry)
+          if old_metadata[:items]
+            new_metadata = Hash[old_metadata[:items].map { |item| [item[:key], item[:value]] }]
```
Temikus, line 111: Consider using `Hash#merge` here: https://ruby-doc.org/core-2.5.1/Hash.html#method-i-merge

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: SGTM. Thank you!

Temikus:
@dvanbrug I think I missed 2 of the earlier comments, so sorry about it. I've resolved all solved threads - can you take a look at the 3 open comments above^

Again, thank you so much for this! :)

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: Perhaps I'm missing something, but here I am converting an Array of Hashes into a Hash.  
```
[{:key => "key1", :value => "value1"}, {:key => "key2", :value => "value2}]

{"key1" => "value1", "key2" => "value2"}
```
My understanding of merge is that it can be used to convert a hash into another hash.  Or perhaps there is a way to use it that I am not thinking of.

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: If you are referring to update_instance_metadata, that method no longer exists.  I think this is from an older commit.

lib/fog/compute/google/requests/reset_windows_password.rb
dvanbrug: I had changed this to be a loop that rechecked every 5 seconds.  After looking at it again i realized it was an infinite loop.  If we never get output here, then the reset operation has failed.  Therefore, I added a loop counter and now raise an exception if we time out after 60 seconds.

dvanbrug:
@Temikus Ok, I've addressed the three comments.  One question, as I've had a lot of changes since the start of the PR, should I squash them all into a single commit?

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: Ooof, my bad. Nope, you're correct here then. Disregard this comment.

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: 👌 Fantastic, thank you!

lib/fog/compute/google/requests/reset_windows_password.rb
Temikus: Ah, I see. Sorry, got confused by the diff here.

Temikus:
@dvanbrug Alright, this looks awesome, thank you so-so much for your contributions and patience :)
Yeah, please squash if you can.

I'll merge this after CI passes.

dvanbrug:
@Temikus Ok, squashed :-).  Thanks again for all your help with this!

Temikus:
@dvanbrug All tests pass, just one tiiiny problem:

```
  test_reset_windows_password                                    ERROR (271.09s)
Minitest::UnexpectedError:         ArgumentError: wrong number of arguments (given 1, expected 0)
            /tmp/build/7e99fb29/src/fog-google/lib/fog/compute/google/models/server.rb:346:in `serial_port_output'
            /tmp/build/7e99fb29/src/fog-google/test/integration/compute/core_compute/test_servers.rb:144:in `test_reset_windows_password'
```

Can you fixup the argument in the method (it's still set to positional):
serial_output = server.serial_port_output(4)

Sorry for the hassle :)

dvanbrug:
Doh!  There we go, fixed up to use the new keyword parameter.

Temikus:
@dvanbrug awesome, merged! I’ll start prepping the release next week.

PR #508, opened 2020-11-23
Fix GoogleXML::File#save incorrectly passing body to headers

### LINKED ISSUE #507: Fog::Storage::GoogleXML::File#save incorrectly passes body to headers
Unlike `Fog::Storage::GoogleJSON::File#save`, `Fog::Storage::GoogleXML::File#save` appears to pass the body to the headers, so this following test script causes a 400 Bad Request:

```ruby
require 'fog/google'

# create a connection
connection = Fog::Storage.new(
{
    provider: 'Google',
    google_project: 'MY-PROJECT',
    google_storage_access_key_id: 'MY-STORAGE-ACCESS-KEY',
    google_storage_secret_access_key: 'MY-STORAGE-SECRET-ACCESS-KEY'
})

dir = connection.directories.new(key: 'MY-BUCKET')

file = dir.files.create(
  key: 'CHANGELOG.md',
  body: "1234\x0000",
)
```

This problem seems to be caused by the implementation of `#metadata`: https://github.com/fog/fog-google/blob/967f89b03403c81947a66007ac08b717c8ad0c24/lib/fog/storage/google_xml/models/file.rb#L54

In this example, `attributes` is this:

```ruby
(byebug) attributes
{:key=>"CHANGELOG.md", :body=>"1234\u000000"}
```

As a result, `body` gets passed through as `options`, so `body` gets used in the headers: https://github.com/fog/fog-google/blob/967f89b03403c81947a66007ac08b717c8ad0c24/lib/fog/storage/google_xml/requests/put_object.rb#L27

I think `metadata` should be using `select` instead of `reject`?

stanhu:
Previously if a file were created with NULL bytes, `File#save` would
fail because the raw body was being set in the header. We now only allow
metadata that starts with `x-goog-meta-` to be passed in as custom
headers.

Closes #507

### DIFF
--- a/test/unit/storage/test_common_xml_collections.rb
+++ b/test/unit/storage/test_common_xml_collections.rb
@@ -6,6 +6,7 @@ class UnitTestStorageXMLCollections < MiniTest::Test
     @client = Fog::Storage.new(provider: "google",
                                google_storage_access_key_id: "",
                                google_storage_secret_access_key: "")
+    @bucket = @client.directories.create(key: "testbucket-#{SecureRandom.hex}")
 
     # Enumerate all descendants of Fog::Collection
     descendants = ObjectSpace.each_object(Fog::Collection.singleton_class)
@@ -35,5 +36,15 @@ class UnitTestStorageXMLCollections < MiniTest::Test
       assert_operator(obj.method(:get).arity, :<=, 1,
                       "#{klass} should have at most 1 required parameter in get()")
     end
+  end
+
+  def test_metadata
+    attr = { key: 'test-file', body: "hello world\x00" }
+    f = @bucket.files.new(attr)
+    assert_equal({}, f.metadata)
+
+    metadata = { 'x-goog-meta-my-header' => 'hello world' }
+    f = @bucket.files.new(attr.merge(metadata))
+    assert_equal(metadata, f.metadata)
   end
 end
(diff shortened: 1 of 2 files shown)


test/unit/storage/test_common_xml_collections.rb
```diff
@@ -36,4 +37,14 @@ def test_collection_get_arguments
                       "#{klass} should have at most 1 required parameter in get()")
     end
   end
+
+  def test_metadata
+    attr = { key: 'test-file', body: "hello world\x00" }
```
stanhu, line 42: This NULL byte doesn't seem to cause the mocked `put_object` to fail on `f.save`. It looks like the mocked version doesn't run:

```ruby
 headers = data[:headers].merge!(options)
``` 

To keep this pull request small, I avoided updating the mocked version to look more like the real version.

Temikus:
@stanhu You're the best, thanks for the PR! Lemme quickly run our main CI and I'll merge if everything's green.

Temikus (commented):
Travis is taking its' sweet time, so I ran the unit tests manually - everything checks out.

Merging.

Temikus (approved):

stanhu:
@Temikus Thanks, would you mind releasing a new version?

Temikus:
@stanhu Yeah, of course. I just wanted to bundle changes from https://github.com/fog/fog-google/pull/509 into it. I expect to roll out today.

PR #517, opened 2020-12-29
Stop using obsoleted URI.encode method

stanhu:
`URI.encode` has been marked obsolete and has been removed from Ruby v3.0.
We use the addressable gem to properly escape the path parameter.

### DIFF
--- a/lib/fog/storage/google_json/utils.rb
+++ b/lib/fog/storage/google_json/utils.rb
@@ -1,3 +1,5 @@
+require 'addressable'
+
 module Fog
   module Storage
     class GoogleJSON
@@ -19,7 +21,7 @@ module Fog
 
         def host_path_query(params, expires)
           params[:headers]["Date"] = expires.to_i
-          params[:path] = URI.encode(params[:path]).gsub("%2F", "/")
+          params[:path] = ::Addressable::URI.encode_component(params[:path], Addressable::URI::CharacterClasses::PATH)
           query = []
 
           if params[:query]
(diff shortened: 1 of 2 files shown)


lib/fog/storage/google_json/utils.rb
```diff
@@ -19,7 +21,7 @@ def url(params, expires)
 
         def host_path_query(params, expires)
           params[:headers]["Date"] = expires.to_i
-          params[:path] = URI.encode(params[:path]).gsub("%2F", "/")
+          params[:path] = ::Addressable::URI.encode_component(params[:path], Addressable::URI::CharacterClasses::PATH).gsub("%2F", "/")
```
stanhu, line 24: I'm not actually sure why `gsub("%2F", "/")` is needed in the first place:

```ruby
irb(main):008:0> URI.encode('first/second/third')
(irb):8: warning: URI.escape is obsolete
=> "first/second/third"
```

stanhu: I suspect this was a legacy change from `CGI.encode`?

```
irb(main):012:0> CGI.escape('first/second/third')
=> "first%2Fsecond%2Fthird"
```

stanhu:
@Temikus Could you take a look at this pull request?

Temikus:
@stanhu I can merge if you can fixup conflicts. Sorry again for the delay.

stanhu:
I think this was resolved in 67c34c3934a5cb0b8063d2bd9f0d496ab16d0e1a, so let's just go with that.

PR #519, opened 2021-03-08
Update README.md with CI issues notice

Temikus:
It really sucks that we have to do this but I see no other way rather than ask for help. I also realise that this org is not owned by us so before I put the notice up I want @plribeiro3000 and @geemus to take a look and see if you folks are ok with such a public plea.

Thank you.

CC @icco


geemus:
I think it's probably fine given the context.

That being said, I know I've loosened my own stance on real tests over time (by necessity). There are quite a few providers that I don't have active accounts on any more. I usually try to mitigate this by using the mock tests, asking folks who do feature work or bug fixes to test it for real, and knowing that I can always rollback changes or releases if I really have to. It's not perfect, but similar to what you are running into here, sometimes there aren't obvious/easy others ways to address it.

geemus (approved):

PR #513, opened 2020-12-13
copy_object: support passing another request options

yosiat:
For example: passing `destination_predefined_acl`

handles: https://github.com/fog/fog-google/issues/505 and https://github.com/fog/fog-google/pull/499#issuecomment-736986810

PR #520, opened 2021-03-14
Support passing object properties to #copy_object

mshibuya:
This PR adds ability to pass [object properties](https://cloud.google.com/storage/docs/json_api/v1/objects?hl=en#resource) like `content_type` on copying Storage objects. At the same time, it also fixes the issue with request_options not being passed properly and having no effect, while taking care of the change in #513 to work as before.

This is needed to fix carrierwaveuploader/carrierwave#2533.

### DIFF
--- a/lib/fog/storage/google_json/requests/copy_object.rb
+++ b/lib/fog/storage/google_json/requests/copy_object.rb
@@ -15,9 +15,19 @@ module Fog
                         target_bucket, target_object, options = {})
           request_options = ::Google::Apis::RequestOptions.default.merge(options)
 
+          object = ::Google::Apis::StorageV1::Object.new(options)
+
           @storage_json.copy_object(source_bucket, source_object,
                                     target_bucket, target_object,
-                                    request_options, **options)
+                                    object, options: request_options, **filter_keyword_args(options))
+        end
+
+        private
+
+        def filter_keyword_args(options)
+          method = @storage_json.method(:copy_object)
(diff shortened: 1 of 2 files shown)


lib/fog/storage/google_json/requests/copy_object.rb
```diff
@@ -15,9 +15,19 @@ def copy_object(source_bucket, source_object,
                         target_bucket, target_object, options = {})
           request_options = ::Google::Apis::RequestOptions.default.merge(options)
 
+          object = ::Google::Apis::StorageV1::Object.new(options)
+
           @storage_json.copy_object(source_bucket, source_object,
                                     target_bucket, target_object,
-                                    request_options, **options)
+                                    object, options: request_options, **filter_keyword_args(options))
```
mshibuya, line 22: `request_options` being passed as 5th argument here was just wrong, it is expected to be a `Google::Apis::StorageV1::Object` according to the google-api-ruby-client.
https://github.com/googleapis/google-api-ruby-client/blob/abb3c968fecb1d62384b75811f20a703bcacd1e7/generated/google-apis-storage_v1/lib/google/apis/storage_v1/service.rb#L1682

ryanrombough:
Bump on this PR to fix https://github.com/carrierwaveuploader/carrierwave/issues/2533

Temikus:
@mshibuya Thanks! :)

Issue #533, opened 2021-06-23
GCS get_object fails with charcode conversion error


mttkay:
Broken out of this discussion: https://github.com/fog/fog-google/pull/530/files#r645936386

With `fog-google` 1.14.0 we are seeing the following error when using the `File.copy` method and the target object is a binary file (in this case, a Java JAR file):

```
Encoding::UndefinedConversionError: "\xC4" from ASCII-8BIT to UTF-8
/opt/gitlab/embedded/lib/ruby/2.7.0/delegate.rb:343:in `write'
/opt/gitlab/embedded/lib/ruby/2.7.0/delegate.rb:343:in `block in delegating_block'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/google-api-client-0.50.0/lib/google/apis/core/download.rb:97:in `block in execute_once'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1093:in `block in follow_redirect'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1250:in `block in do_get_block'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient/session.rb:894:in `read_body_length'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient/session.rb:587:in `get_body'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1247:in `do_get_block'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1019:in `block in do_request'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1133:in `protect_keep_alive_disconnected'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1014:in `do_request'
/opt/gitlab/embedded/lib/ruby/gems/2.7.0/gems/httpclient-2.8.3/lib/httpclient.rb:1104:in `follow_redirect'
[...]

PR #534, opened 2021-06-23
Fix get_object not working with binary files

stanhu:
The change in #521 caused the `Tempfile` buffer to be used, which opens
a file in ASCII mode by default. We need to make use of `binmode` to
activate binary streams.

Closes https://github.com/fog/fog-google/issues/533

### DIFF
--- a/test/integration/storage/test_objects.rb
+++ b/test/integration/storage/test_objects.rb
@@ -34,13 +34,15 @@ class TestStorageRequests < StorageShared
 
   def test_put_object_paperclip
     object_name = new_object_name
-    paperclip_file = OpenStruct.new(:path => some_temp_file,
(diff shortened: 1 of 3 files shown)


stanhu:
@geemus @Temikus Would you mind reviewing this? This is a pretty serious bug.

test/integration/storage/test_objects.rb
```diff
@@ -41,6 +41,7 @@ def test_put_object_paperclip
     object = @client.get_object(some_bucket_name, object_name)
 
     assert_equal(object_name, object[:name])
+    assert_equal(Encoding::ASCII_8BIT, object[:body].encoding)
```
stanhu, line 44: Let me improve this test by adding binary data.

geemus (approved):
Looks good, thanks for the quick fix!

geemus:
Fix released in v1.15.0.

PR #539, opened 2021-07-17
Add self-hosted runners to Fog-google

Temikus:


.github/workflows/ruby.yml
```diff
@@ -32,3 +31,28 @@ jobs:
       run: bundle install
     - name: Run tests
       run: bundle exec rake
+
+  test-integrate:
+    runs-on: self-hosted
+    strategy:
+      matrix:
+        ruby-version: ['2.7']
```
icco, line 39: We should probably test 3.0 too

Temikus: Yeah, we'll set that up - this is just me trying on one thing.

Issue #541, opened 2021-07-20
[storage] File `#body` can download the file twice?


10io:
Hello! 👋 

I've been using this adapter to access files on GCP Storage and hit an issue.
If I head the file and then I access `#body`, the file seems to be downloaded twice.
This is ok for small files but can quickly become a problem for bigger files.

Here is how to reproduce this:
```ruby
storage = ::Fog::Storage.new(credentials) # `credentials` is the standard hash with the parameters to connect to Google Storage
directory = storage.directories.new(key: directory) # `directory` is the bucket name
f = directory.files.head(path) # `path` is the path for the file within the bucket
f.body
```

Here is the log (truncated) from `google-api-client` when `f.body` is executed:

```
Sending HTTP get https://storage.googleapis.com/storage/[object_path]
200
#<HTTP::Message:0x00007fd79d55d170 >
Success - #<Google::Apis::StorageV1::Object:0x00007fd7ac088ff0 >

Success - #<File:[temporary_folder]/fog-google-storage-temp20210720-69390-1uveobl>

Sending HTTP get https://storage.googleapis.com/storage/[object_path]
200
#<HTTP::Message:0x00007fd79d514330 >
Success - #<Google::Apis::StorageV1::Object:0x00007fd79d504980 >

Success - #<File:[temporary_folder]/fog-google-storage-temp20210720-69390-zrp5w8>
```

I see two temporary files being created. I put some additional logs around those files and I can confirm that both have the same size and the same content. To me, the file really seems to be downloaded twice. 😿 In addition, those `Sending HTTP get` lines seems to indicate that the API is hit twice.

From my code inspection, the problem seems to come from https://github.com/fog/fog-google/blob/98a00e1bbace13a65b42f1b83c1911ac01ce71c0/lib/fog/storage/google_json/models/file.rb#L48
If I get this correctly, `collection.get(identity)` and `file.body` both download the file. Is this expected?

I don't fully grasp that line of code: I don't get why we need `collection.get(identity)`. I assume that it has something to do with a `Fog::Storage::GoogleJSON::File` coming from a metadata [request](https://cloud.google.com/storage/docs/json_api/v1/objects/get) vs a `Fog::Storage::GoogleJSON::File` coming from a "full" [get (with the `alt=media` query parameter)](https://cloud.google.com/storage/docs/json_api/v1/objects/get).

stanhu:
[...]

PR #545, opened 2021-08-03
Avoid duplicate GET requests when retrieving body

stanhu:
Previously the response body would be loaded twice in memory whenever
`GoogleJSON::File#body` was called. This was happening because inside
the ternary condition, `file.body` was recursively calling itself twice
even though the body was already loaded. We rewrite this method to be
simpler and to have a proper base condition.

Closes https://github.com/fog/fog-google/issues/541

### DIFF
--- a/lib/fog/storage/google_json/models/file.rb
+++ b/lib/fog/storage/google_json/models/file.rb
@@ -45,7 +45,16 @@ module Fog
         end
 
         def body
-          last_modified && (file = collection.get(identity)) ? attributes[:body] ||= file.body : attributes[:body] ||= ""
+          return attributes[:body] if attributes.key?(:body)
+
+          file = collection.get(identity)
+
+          attributes[:body] =
+            if file
+              file.attributes[:body]
+            else
+              ""
+            end
         end
 
         def body=(new_body)


lib/fog/storage/google_json/models/file.rb
```diff
@@ -45,7 +45,16 @@ def predefined_acl=(new_predefined_acl)
...
 
         def body
-          last_modified && (file = collection.get(identity)) ? attributes[:body] ||= file.body : attributes[:body] ||= ""
+          return attributes[:body] if attributes.key?(:body)
+
+          file = collection.get(identity)
+
+          attributes[:body] =
+            if file
+              attributes[:body]
```
stanhu, line 54: This is not quite right.

lib/fog/storage/google_json/models/file.rb
```diff
@@ -45,7 +45,16 @@ def predefined_acl=(new_predefined_acl)
...
 
         def body
-          last_modified && (file = collection.get(identity)) ? attributes[:body] ||= file.body : attributes[:body] ||= ""
+          return attributes[:body] if attributes.key?(:body)
+
+          file = collection.get(identity)
+
+          attributes[:body] =
+            if file
+              file.attributes[:body]
```
stanhu, line 54: I opted to do this and ditch `last_modified` to avoid any potential of an infinite recursion. I'm not quite sure what `last_modified` has to do with this in any case.

Temikus:
@stanhu Thank you 🙌 I'll run the integration and then merge.

Temikus:
Storage passing - merging. Still need to work out some kinks in flaky tests 🤔

stanhu:
@Temikus Would you mind releasing a new version?

Temikus:
@stanhu will kick off today, sorry for delay.

PR #550, opened 2021-09-30
Add support for instance display device

dcode:
Admittedly, I'm not 100% confident in this change, but it looks simple enough. This PR adds support for the [:display_device](https://googleapis.dev/ruby/google-api-client/v0.39.2/Google/Apis/ComputeV1/Instance.html#display_device-instance_method) option on a ComputeInstance.

I think one would use it like so (adapted from https://github.com/mitchellh/vagrant-google/, which is ultimately where I'd like to use it):

```ruby
            display_device = { :enable_display => true }

            if display_device.has_value?(true)
              defaults[:display_device] = display_device
            end

            # Assume `defaults` is populated with the bare minimum needed values for the `options` parameter
            server = env[:google_compute].servers.create(defaults)
```


dcode:
Okay, I at least tested this with manual edits to my local vagrant environment and it works. Ruby just isn't my strongest programming language and this was freehand edited.... but it at least works. Let me know if there's some style or idiom concepts that I missed.

Temikus:
@dcode Thank you for your contribution!

As an addition to this change you also need to add an attribute to the `Server` model: https://github.com/fog/fog-google/blob/master/lib/fog/compute/google/models/server.rb

You can use this PR as an example - it changes the `Disk` model but it's very similar: https://github.com/fog/fog-google/commit/443c233759e3bf2a196d61124c84c5cafaa28516

Issue #323, opened 2018-04-17
"Google::Apis::TransmissionError: execution expired" on large files


mathportillo:
Hello,
First, thanks a lot for this gem!
I am getting the above error when uploading large files (1GB)
I am trying to find a way to define timeout seconds for write operations.
I've searched everywhere on fog repo, paperclip repo, google-api-ruby-client repo, and this repo.

Here is the stack trace (gem versions are shown with it):
```
Caused by:
Google::Apis::TransmissionError: execution expired
[...]/httpclient-2.8.3/lib/httpclient/ssl_socket.rb:80:in `<<'
[...]/httpclient-2.8.3/lib/httpclient/http.rb:582:in `dump_file'
[...]/httpclient-2.8.3/lib/httpclient/http.rb:508:in `dump'
[...]/httpclient-2.8.3/lib/httpclient/http.rb:962:in `dump'
[...]/httpclient-2.8.3/lib/httpclient/session.rb:517:in `block in query'
[...]/httpclient-2.8.3/lib/httpclient/session.rb:515:in `query'
[...]/httpclient-2.8.3/lib/httpclient/session.rb:177:in `query'
[...]/httpclient-2.8.3/lib/httpclient.rb:1242:in `do_get_block'
[...]/httpclient-2.8.3/lib/httpclient.rb:1019:in `block in do_request'
[...]/httpclient-2.8.3/lib/httpclient.rb:1133:in `protect_keep_alive_disconnected'
[...]/httpclient-2.8.3/lib/httpclient.rb:1014:in `do_request'
[...]/httpclient-2.8.3/lib/httpclient.rb:1104:in `follow_redirect'
[...]/httpclient-2.8.3/lib/httpclient.rb:854:in `request'
[...]/httpclient-2.8.3/lib/httpclient.rb:765:in `post'
[...]/google-api-client-0.19.8/lib/google/apis/core/upload.rb:228:in `send_upload_command'
[...]/google-api-client-0.19.8/lib/google/apis/core/upload.rb:254:in `execute_once'
[...]/google-api-client-0.19.8/lib/google/apis/core/http_command.rb:104:in `block (2 levels) in execute'
[...]/retriable-3.1.1/lib/retriable.rb:61:in `block in retriable'
[...]/retriable-3.1.1/lib/retriable.rb:57:in `times'
[...]/retriable-3.1.1/lib/retriable.rb:57:in `retriable'
[...]

PR #560, opened 2022-01-24
Disable transfer timeout by default in storage request

stanhu:
By default, httpclient will prevent any request from going longer than 120 seconds, which will cause any transfer to take longer than that to abort. As a result, the Google client will raise `Google::Apis::TransmissionError: execution expired` for any transfer that is longer than 2 minutes.

`Google::Apis::RequestOptions.default.retries` can be used to resume aborted uploads, but the default is set to 0. Plus, `max_elapsed_time` is set to 15 minutes by default, so this value would need to be raised as well. Support for this was added in https://github.com/googleapis/google-api-ruby-client/pull/8106, but this is a recent feature.

The main way we can fix this issue is to raise the timeout to a large value to disable it outright. Since httpclient uses
`Timeout::timeout`, which can be dangerous, we disable this timeout by default if no `send_timeout_sec` is specified.

Closes https://github.com/fog/fog-google/issues/323

### DIFF
--- a/lib/fog/storage/google_json/real.rb
+++ b/lib/fog/storage/google_json/real.rb
@@ -26,7 +26,7 @@ module Fog
 
           @storage_json.client_options.open_timeout_sec = options[:open_timeout_sec] if options[:open_timeout_sec]
           @storage_json.client_options.read_timeout_sec = options[:read_timeout_sec] if options[:read_timeout_sec]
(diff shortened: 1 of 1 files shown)


Temikus:
@stanhu disabling timeouts are generally dangerous, what prevents someone from setting that timeout manually when they need to upload large files?

I feel I'm missing something here 🤔

PR #580, opened 2022-10-24
Update deprecated images & image families

Temikus:
Mainly fixing failing tests due to expired images and image families.


Temikus:
Aight, merging this in, tests look much better, but the monitoring ones need more work, will address in separate PR.

PR #582, opened 2022-10-31
Attempt to make integration tests lighter

Temikus:
Starting with separating SQL integration tests into their own job as that allows us to only trigger it on PR's where the code or deps have actually changed.


Temikus:
Tested - workflows trigger correctly 👍

PR #583, opened 2022-10-31
Upgrade google client library dependencies

Temikus:
Upgrading Google Library dependencies across the board, it's been a while so will wait for a full integration test roll.


Temikus:
All checks passed, merged 👍

Issue #590, opened 2023-03-16
Issue with escaped URLs with plus sign


ihollander:
We ran into an issue after updating to the latest version of `fog-google` where filenames with a `+` character in them are escaped differently in URLs than they were prior to updating. 

Prior to https://github.com/fog/fog-google/commit/67c34c3934a5cb0b8063d2bd9f0d496ab16d0e1a, when generating a URL using fog-google, given a file with a path like this, we'd end up with a URL like this (we use fog-google via Carrierwave):

```rb
model.uploaded_file.path
# => "path/to/test_upload+.pdf"
model.uploaded_file.url
# => "https://storage.googleapis.com/bucket/path/to/test_upload+.pdf?GoogleAccessId=..."
```

After updating fog-google, the same file has `+` signs in its path encoded as `%2B`:

```rb
model.uploaded_file.path
# => "path/to/test_upload+.pdf"
model.uploaded_file.url
# => "https://storage.googleapis.com/bucket/path/to/test_upload%2B.pdf?GoogleAccessId=..."
```

It seems this change causes issues when we hand off this URL to ngnix, which we use to proxy requests for files in our GCP buckets:

```rb
def stream_from_cloud_storage(url)
  headers['X-Accel-Redirect'] = url
  # ...
end
```

When this happens, the behavior I see is that nginx handles these redirect correctly for URLs without `%2B` in the path, but doesn't handle this correctly when the URL has `%2B` in the path (other encoded characters work fine - it's really just `+` -> `%2B` that causes issues).

It's not totally clear to me what is causing this issue with nginx redirects specifically (I think it may be changing the path in the URI and causing the signature in the query params to not match the path), but the only solution I could find that lets us preserve our current nginx setup is to stop fog-google from escaping `+` signs: https://github.com/ihollander/fog-google/commit/68b7d699979af6a0510f9abe0ad3aaf60ecc2d3d

I understand the change to remove the old implementation of `URI.escape` in https://github.com/fog/fog-google/commit/67c34c3934a5cb0b8063d2bd9f0d496ab16d0e1a is necessary because `URI.escape` is obsolete, but there was another approach to this discussed in https://github.com/fog/fog-google/pull/517 using the `Addressable` library, which may be more appropriate for handling these URIs than the `CGI.escape`-inspired approach (and gives results are more similar to the old `URI.escape` approach). For comparison, here's how certain characters are escaped by each of these approaches:

```rb
# adapted from https://gist.github.com/rcarver/66498
[...]

PR #595, opened 2023-06-12
Show error when filename has `+` in it.

mobilutz:
This shows that the error still exists in current `fog-google` implementation.

Ref: https://github.com/fog/fog-google/issues/590


stanhu:
I think we can close this in favor of https://github.com/fog/fog-google/pull/596.

geemus:
Closing in favor of #596

Issue #607, opened 2023-09-27
Unable to list more than 500 instances with `Fog::Compute::Google::Servers#all`


agrare:
If there are more results than will fit in a page (default size 500 but can be limited with `:max_results`) then a `nextPageToken` is returned with the `InstanceList`.

The [`Fog::Compute::Google::Servers#all`](https://github.com/fog/fog-google/blob/v1.22.0/lib/fog/compute/google/models/servers.rb#L7-L27) method ignores this and returns the items in the first page only.

There is an optional `:page_token` parameter to `#all` but the next_page_token isn't returned to the caller so it isn't obvious how this was intended to be used.

I don't personally have an environment with this many instances but the issue can be easily demonstrated by using `:max_results` to force `list_servers` and `list_aggregated_servers` to return a smaller page with a `nextPageToken`:
```
>> compute.servers.all.count
=> 4
>> compute.servers.all(:max_results => 1).count
=> 1
```

In this case `max_results` limits `service.list_aggregated_servers` to return a single server plus a nextPageToken which is ignored.  If the page size were not arbitrarily limited then only the first 500 instances would be returned.

This is fundamentally the same issue as https://github.com/fog/fog-google/issues/221 and https://github.com/fog/fog-google/issues/42

I believe all of the `Model#all` methods likely have this same issue.

PR #608, opened 2023-09-27
Fix `Fog::Compute::Google::Servers#all` paging

agrare:
If there are more instances than will fit in a page a next_page_token is returned.  This has to be passed in to subsequent calls in order to retrieve the full list of servers.

Note: I think this method could use some refactoring but I wanted to model this after the way `Snapshots#all` was fixed for consistency and change as little as possible for a targeted bug fix.

Fixes https://github.com/fog/fog-google/issues/607

### DIFF
--- a/lib/fog/compute/google/models/servers.rb
+++ b/lib/fog/compute/google/models/servers.rb
@@ -13,17 +13,31 @@ module Fog
             :page_token => page_token
           }
 
-          if zone
-            data = service.list_servers(zone, **opts).to_h[:items] || []
-          else
-            data = []
-            service.list_aggregated_servers(**opts).items.each_value do |scoped_lst|
-              if scoped_lst && scoped_lst.instances
-                data.concat(scoped_lst.instances.map(&:to_h))
+          items = []
+          next_page_token = nil
+
+          loop do
+            if zone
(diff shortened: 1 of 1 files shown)


agrare:
I'm curious on the maintainers' points of view on this but options to this method include `:max_results` and conceptually it isn't clear if this is intended to mean "never return more than this number of results" or specifically meant to be passed directly to the `list_servers` and `list_aggregated_servers` methods.

These options were added here https://github.com/fog/fog-google/commit/9891535edbd546247916cbaed3d9349dbcdb24d3#diff-bb76744c2fc47cf4c93fd72cfe6563ceb8d1421a342dd36a583fac4e49178142R7-R8 in what looks like simply adding explicit kwargs for each of the options that `list_servers` and `list_aggregated_servers` take.

I could see a user wanting to limit both the total results and the size of each page so maybe an additional option is needed?

agrare:
cc @Temikus

Temikus (approved):
LGTM

Regarding semantics - that particular commit was during a massive quick-and-dirty rewrite when google-cloud lib broke compatibility, so I'm in favor of a refactor to whatever is more user-friendly.

PR #629, opened 2024-04-09
storage: fix incorrect API scopes for IAM SignBlob API

### LINKED ISSUE #599: Fog::Storage::GoogleJSON::Real#iam_signer is using the wrong API scopes with Workload Identity
I was investigating why running GitLab in Kubernetes with Workload Identity instead of a service account key resulted in `ACCESS_TOKEN_SCOPE_INSUFFICIENT` errors:

```ruby
irb(main):045:0> credentials = {provider: 'Google', google_project: 'my-gitlab-project', google_application_default: true}
=> {:provider=>"Google", :google_project=>"my-gitlab-project", :google_application_default=>true}
irb(main):046:0> raw_config = {enabled: true, connection: credentials, remote_directory: 'my-gitlab-project-uploads', storage_options: {}, consolidated_settings: false}
=>
{:enabled=>true,
...
irb(main):047:0> config = ObjectStorage::Config.new(raw_config)
=>
#<ObjectStorage::Config:0x00007f3d35837538
...
irb(main):048:0> du = ObjectStorage::DirectUpload.new(config, 'tmp/foo/bar', has_length: true)
=>
#<ObjectStorage::DirectUpload:0x00007f3d35fdbc58
...
irb(main):049:0> du.get_url rescue puts $!.body
{
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "ACCESS_TOKEN_SCOPE_INSUFFICIENT",
        "domain": "googleapis.com",
        "metadata": {
          "method": "google.iam.credentials.v1.IAMCredentials.SignBlob",
          "service": "iamcredentials.googleapis.com"
        }
      }
    ]
  }
}
=> nil
```

Digging into `fog-google` I have found 2 issues causing this:

1. `Fog::Storage::GoogleJSON::Real#iam_signer` is using the wrong API scope, it should be `https://www.googleapis.com/auth/iam` instead of `https://www.googleapis.com/auth/devstorage.full_control`: https://github.com/fog/fog-google/blob/30685949ff7688e1066e4c8625480caffc7a495f/lib/fog/storage/google_json.rb#L32
2. Fixing the API scope above alone is not enough, as `Fog::Storage::GoogleJSON::Real#iam_signer` seems to be re-using the same session and auth credentials as with the storage operations, I haven't been able to find where and why yet but changing `GOOGLE_STORAGE_JSON_API_SCOPE_URLS`  to `https://www.googleapis.com/auth/cloud-platform` did indeed make it work, though that's not the ideal solution

stanhu:
Previously when a service account attempted to use the IAM SignBlob API, the request would fail with a 403
`ACCESS_TOKEN_SCOPE_INSUFFICIENT` because the wrong scope was requested.

As documented in
https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob, either `https://www.googleapis.com/auth/iam` or
`https://www.googleapis.com/auth/cloud-platform` is needed.

This commit fixes an issue where the default authorization header with the `https://www.googleapis.com/auth/devstorage.full_control` scope was being used by the IAM service. This occurred because the previous code did not actually set the scope properly, and for the IAM service to work properly, we need to request a new access token with the correct scope.

Note that the service account in question needs to have the `Service Account Token Creator` IAM role to work.

Closes #599

### DIFF
--- a/lib/fog/storage/google_json.rb
+++ b/lib/fog/storage/google_json.rb
@@ -29,7 +29,7 @@ module Fog
 
       # Version of IAM API used for blob signing, see Fog::Storage::GoogleJSON::Real#iam_signer
       GOOGLE_STORAGE_JSON_IAM_API_VERSION = "v1".freeze
-      GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS = %w(https://www.googleapis.com/auth/devstorage.full_control).freeze
+      GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS = %w(https://www.googleapis.com/auth/iam).freeze
 
       # TODO: Come up with a way to only request a subset of permissions.
       # https://cloud.google.com/storage/docs/json_api/v1/how-tos/authorizing
--- a/lib/fog/storage/google_json/real.rb
+++ b/lib/fog/storage/google_json/real.rb
@@ -10,16 +10,12 @@ module Fog
 
         def initialize(options = {})
           shared_initialize(options[:google_project], GOOGLE_STORAGE_JSON_API_VERSION, GOOGLE_STORAGE_JSON_BASE_URL)
+          @options = options.dup
           options[:google_api_scope_url] = GOOGLE_STORAGE_JSON_API_SCOPE_URLS.join(" ")
           @host = options[:host] || "storage.googleapis.com"
 
           # TODO(temikus): Do we even need this client?
           @client = initialize_google_client(options)
-          # IAM client used for SignBlob API
-          @iam_service = ::Google::Apis::IamcredentialsV1::IAMCredentialsService.new
-          apply_client_options(@iam_service, {
-                                 google_api_scope_url: GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS.join(" ")
-                               })
 
           @storage_json = ::Google::Apis::StorageV1::StorageService.new
           apply_client_options(@storage_json, options)
@@ -141,6 +137,18 @@ DATA
           return key.sign(digest, string_to_sign)
         end
 
+        # IAM client used for SignBlob API.
+        # Lazily initialize this since it requires another authorization request.
+        def iam_service
+          return @iam_service if defined?(@iam_service)
+
+          @iam_service = ::Google::Apis::IamcredentialsV1::IAMCredentialsService.new
+          apply_client_options(@iam_service, @options)
+          iam_options = @options.merge(google_api_scope_url: GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS.join(" "))
+          @iam_service.authorization = initialize_auth(iam_options)
+          @iam_service
+        end
+
         ##
         # Fallback URL signer using the IAM SignServiceAccountBlob API, see
         #   Google::Apis::IamcredentialsV1::IAMCredentialsService#sign_service_account_blob
@@ -162,7 +170,7 @@ DATA
           )
 
           resource = "projects/-/serviceAccounts/#{google_access_id}"
-          response = @iam_service.sign_service_account_blob(resource, request)
+          response = iam_service.sign_service_account_blob(resource, request)
 
           return response.signed_blob
         end
--- a/lib/fog/google/shared.rb
+++ b/lib/fog/google/shared.rb
@@ -67,20 +67,21 @@ module Fog
           ::Google::Apis.logger.level = ::Logger::DEBUG
         end
 
-        auth = nil
-
-        if options[:google_json_key_location] || options[:google_json_key_string]
-          auth = process_key_auth(options)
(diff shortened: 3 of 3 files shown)


lib/fog/storage/google_json/real.rb
```diff
@@ -10,16 +10,12 @@ class Real
 
         def initialize(options = {})
           shared_initialize(options[:google_project], GOOGLE_STORAGE_JSON_API_VERSION, GOOGLE_STORAGE_JSON_BASE_URL)
+          @options = options.dup
           options[:google_api_scope_url] = GOOGLE_STORAGE_JSON_API_SCOPE_URLS.join(" ")
           @host = options[:host] || "storage.googleapis.com"
 
           # TODO(temikus): Do we even need this client?
```
stanhu, line 17: I was tempted to drop this and `@client` since they're not actually used.

lib/fog/storage/google_json/real.rb
```diff
@@ -141,6 +137,18 @@ def default_signer(string_to_sign)
...
         end
 
+        # IAM client used for SignBlob API.
+        # Lazily initialize this since it requires another authorization request.
+        def iam_service
+          return @iam_service if defined?(@iam_service)
+
+          @iam_service = ::Google::Apis::IamcredentialsV1::IAMCredentialsService.new
+          apply_client_options(@iam_service, @options)
+          iam_options = @options.merge(google_api_scope_url: GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS.join(" ")))
```
stanhu, line 147: In the future, it might be nicer to distinguish between options that are only used by this gem and options that are used by the Google SDK. It's not obvious that `google_api_scope_url` isn't passed all the way to the Google SDK.

stanhu:
@Temikus Would you mind reviewing this? This seems to be a pretty critical fix for Google Kubernetes Engine users.

lib/fog/storage/google_json.rb
```diff
@@ -29,7 +29,7 @@ class GoogleJSON < Fog::Service
 
       # Version of IAM API used for blob signing, see Fog::Storage::GoogleJSON::Real#iam_signer
       GOOGLE_STORAGE_JSON_IAM_API_VERSION = "v1".freeze
-      GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS = %w(https://www.googleapis.com/auth/devstorage.full_control).freeze
+      GOOGLE_STORAGE_JSON_IAM_API_SCOPE_URLS = %w(https://www.googleapis.com/auth/iam).freeze
```
stanhu, line 32: @peikk0 Should we use https://www.googleapis.com/auth/cloud-platform instead?

stanhu: https://www.googleapis.com/auth/cloud-platform seems much more permissive. I'm inclined to stick with this?

peikk0: On one hand, `https://www.googleapis.com/auth/iam` is more restrictive and keeps the original intent. On the other hand, access scopes are considered legacy by Google and they recommend using `https://www.googleapis.com/auth/cloud-platform` and proper IAM permissions instead: https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam But changing to that could have security implications for users that still rely on it so it might be safer to stick to `https://www.googleapis.com/auth/iam` for now.

stanhu: 👍 Thanks!

lib/fog/storage/google_json/real.rb
Temikus: You use this API more than me so I am happy to drop this if there are no corner cases here 👍

Temikus (approved):
LGTM - feel free to drop the client too

Temikus:
And thanks for your contributions @stanhu as usual ❤️

lib/fog/storage/google_json/real.rb
stanhu: Let's leave this in for now.

Temikus:
Kicked off the CI, once it passes will merge and release as soon as I can.

Temikus:
@stanhu looks like there’s a small typo causing some errors - you can see the unit test run failing ‘rake test:unit’ for local if you need it.

stanhu:
> @stanhu looks like there’s a small typo causing some errors - you can see the unit test run failing ‘rake test:unit’ for local if you need it.

Thanks. That was a last-minute refactor, should be fixed now.

Temikus:
LGTM, merging 👍 Will aim to release today if I can - if not - will push Friday.

Issue #639, opened 2025-06-03
fog-google modifies strings


dgmora:
I've tried to enable  `RUBYOPT="--enable-frozen-string-literal"` so that strings are frozen for my application, but I noticed that fog-google modifies frozen strings:

```
FrozenError: can't modify frozen String: "/"
storage_json/real.rb:52:in 'Fog::Google::StorageJSON::Real#signature'
```

Strings will be frozen eventually for Ruby, so it would be great if fog-google would be prepared for that.

I am not sure in how many places fog-google modifies strings. Adding `# frozen_string_literal: true` / `# frozen_string_literal: false` could already help people using `--enable-frozen-string-literal`, or maybe these cases could be fixed directly

PR #640, opened 2025-06-04
Add `frozen_string_literal: false` to files that modify strings and run tests with frozen strings

davidgm0:
Partially addresses #639. It runs tests with frozen strings, but allows modifying strings for those files with `frozen_string_literal: false`. This should allow people using for to run with `RUBYOPT="--enable-frozen-string-literal"`. Next step would involve not modifying strings, but this was more involved, so I thought it would make sense to do this first.

### DIFF
--- a/.github/workflows/integration-storage.yml
+++ b/.github/workflows/integration-storage.yml
@@ -36,6 +36,8 @@ concurrency:
 jobs:
   test:
     runs-on: fog-arc-runner
+    env:
+      RUBYOPT: "--enable-frozen-string-literal"
     strategy:
       matrix:
         ruby-version: [ '3.0', '3.1', '3.2' ]
--- a/.github/workflows/integration-compute-core.yml
+++ b/.github/workflows/integration-compute-core.yml
@@ -34,6 +34,8 @@ concurrency:
 jobs:
   test-compute:
     runs-on: fog-arc-runner
+    env:
+      RUBYOPT: "--enable-frozen-string-literal"
     strategy:
       matrix:
         ruby-version: [ '3.0', '3.1', '3.2' ]
--- a/.github/workflows/integration-compute-instance_groups.yml
+++ b/.github/workflows/integration-compute-instance_groups.yml
@@ -35,6 +35,8 @@ concurrency:
 jobs:
   test:
     runs-on: fog-arc-runner
+    env:
+      RUBYOPT: "--enable-frozen-string-literal"
     strategy:
       matrix:
         ruby-version: [ '3.0', '3.1', '3.2' ]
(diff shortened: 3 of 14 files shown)


.github/workflows/integration-storage.yml
```diff
@@ -36,6 +36,8 @@ concurrency:
 jobs:
   test:
     runs-on: fog-arc-runner
+    env:
+      RUBYOPT: "--enable-frozen-string-literal"
```
stanhu, line 40: Wouldn't we expect these tests to fail without fixes for the frozen strings?

davidgm0: To those files that modify strings I added:
# frozen_string_literal: **false** 

That overrides the `RUBYOPT` for those files. The final behaviour should be that by default for all files the strings are frozen and for those few files the string literals are not frozen.

stanhu: Oh, right, that's unexpected, though; usually I see `# frozen_string_literal: true`, or it's omitted entirely because `false` is the default.

I'm not opposed to this approach as a step 1. However, at the moment these integration tests need to be run by hand. I suspect just fixing the frozen string issue is relatively straightforward that we might as well try to fix them on a file-by-file basis, such as:

```diff
diff --git a/lib/fog/google/storage/storage_json/real.rb b/lib/fog/google/storage/storage_json/real.rb
index 23c86c5..ab99372 100644
--- a/lib/fog/google/storage/storage_json/real.rb
+++ b/lib/fog/google/storage/storage_json/real.rb
@@ -45,7 +45,7 @@ DATA
           end
           string_to_sign << canonical_google_headers.to_s
 
-          canonical_resource = "/"
+          canonical_resource = +"/"
           if subdomain = params.delete(:subdomain)
             canonical_resource << "#{CGI.escape(subdomain).downcase}/"
           end
diff --git a/lib/fog/google/storage/storage_xml/real.rb b/lib/fog/google/storage/storage_xml/real.rb
index 166e181..5c15bfc 100644
--- a/lib/fog/google/storage/storage_xml/real.rb
+++ b/lib/fog/google/storage/storage_xml/real.rb
@@ -57,7 +57,7 @@ DATA
           end
           string_to_sign << canonical_google_headers.to_s
 
-          canonical_resource = "/"
+          canonical_resource = +"/"
           if subdomain = params.delete(:subdomain)
             canonical_resource << "#{CGI.escape(subdomain).downcase}/"
           end
```

dgmora: I'd be happy if these changes could be applied directly, I just don't know the codebase so well and I saw there were a fair amount of modifications, so I went for the easy fix. That can still be applied after this PR.

> Oh, right, that's unexpected, though; usually I see # frozen_string_literal: true, or it's omitted entirely because false is the default.

Yeah, but that default will eventually change. If someone tries to already enable `--enable-frozen-string-literal` (as I did), they will get errors. This PR at least fixes that

stanhu:
Let's close this in favor of https://github.com/fog/fog-google/pull/646.

PR #657, opened 2026-02-09
Fix handling of universe domains with service key

stanhu:
The Google API already extracts the universe domain from the service key, if available. We don't need to set it.

Note that we still need to set the universe domain in client options because the Ruby client does not yet do this automatically unless you set the `disable_universe_domain_check` flag to false due to https://github.com/googleapis/google-auth-library-ruby/pull/493:

```
require 'googleauth'
creds = Google::Auth.get_application_default
creds.disable_universe_domain_check = false
creds.fetch_access_token!
```


geemus:
Thanks!

PR #660, opened 2026-09-13
Keep options given to StorageJSON::Files#all for subsequent pages

mshibuya:
## Problem

`StorageJSON::Files#all` passes its options to `list_objects` without keeping them on the collection. `#each` lists again from a copy (`dup.all`), so every page after the first is listed without them:

```ruby
directory.files.all(prefix: "a/").each { |file| ... } # iterates over the whole bucket
```

Anything that deletes what it iterates over will delete objects outside the prefix. We found this through CarrierWave's cache cleanup (`directory.files.all(prefix: cache_dir).each { |file| file.destroy if ... }`): with HTTP stubbed, it listed the second page onward without the prefix and sent DELETEs for objects outside the cache directory.

The XML, AWS, Aliyun and AzureRM collections already call `merge_attributes(options)` in `#all`, so only the JSON storage is affected.

## Change

`#all` merges the options into the collection's attributes before listing, as the other providers do. As a side effect, options can also be given by attribute name or alias (`"prefix"`, `"Prefix"`); these used to be dropped silently.

Options now stay on the collection: `all(page_token: x)` followed by `all` continues from `x` instead of starting over, as `marker` already does for AWS and the XML storage.

## Tests

- `test/unit/storage/test_json_files.rb`: every page is listed with the prefix, and `"prefix"`, `"Prefix"` and `:prefix` are all accepted. `list_objects` is stubbed, since the JSON mock doesn't implement it.
- `test/unit/storage/test_xml_files.rb`: the XML storage already behaves this way but wasn't tested.
- `test/integration/storage/test_files.rb`: iterates `all(prefix: "a/", max_results: 2)` over two pages against the real API.

The JSON unit test and the integration test fail without the fix. `rake test:unit` passes, and the new integration test passes against GCS with application default credentials (I didn't run the rest of the storage suite).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


geemus (approved):
lgtm, thanks!

geemus:
Something seems broken with integration tests (they hung for >24hrs), but this seems clear/safe, so bringing it in.

